PR #129, opened 2024-02-08
Add e2e test for tail sampling config example

joaopgrassi:
After spending some time learning/looking at the [testbed](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/testbed) package in the collector, here's what I came up with to run e2e tests for our configuration samples:

- New data Provider (`sampleConfigsDataProvider`): This implements the [DataProvider](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/testbed/testbed/data_providers.go#L24) interface. Since we have samples with different expectation/results, the data provider simply receives a list of spans, and when the load generator asks for it, it returns that. This way, each test is in full control of the data that is sent to the collector

- New validator `(SampleConfigsValidator`): This implements the [TestCaseValidator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/testbed/testbed/validator.go#L20) interface, and the same way the data provider, receives the expected spans. It then compares the expected with the received from the mock back-end. For now the assertion simply looks for SpanIDs.

Then it's just using the testbed framework to actually run the collector/tests. For now, I'm testing the tail sampling configuration.


config_examples/tail_sampling.yaml
```diff
@@ -19,14 +19,19 @@ processors:
       - name: keep-slow-traces
         type: latency
         latency: {threshold_ms: 500}
+    decision_wait: 30s
```
joaopgrassi, line 22: This is the default value, but I had to add to be able to easily string replace it for the test. This interval is too long and makes the test timeout, as the default `go test` timeout is also 30s. Let me know if you have a better idea for this.

testbed/integration/data_providers.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"sync/atomic"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/testbed/testbed"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+// sampleConfigsDataProvider in an implementation of the DataProvider
```
evan-bradley, line 12:
```suggestion
var _ testbed.DataProvider = &sampleConfigsDataProvider{}

```

I forget what these statements are called, but I would just verify that the type implements the interface like this. This also allows you to remove the comments on the methods since you can see which methods the interface provides.

config_examples/tail_sampling.yaml
joaopgrassi: Yep agreed. I will leave it like this then.

testbed/integration/config_examples_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	receiverPort := testbed.GetAvailablePort(t)
+	exporterPort := testbed.GetAvailablePort(t)
+
+	parsedConfig := string(cfg)
+	parsedConfig = replaceOtlpGrpcReceiverPort(parsedConfig, receiverPort)
+	parsedConfig = replaceDynatraceExporterEndpoint(parsedConfig, exporterPort)
+
+	// replaces the sampling decision wait so the test doesn't timeout
+	parsedConfig = strings.Replace(parsedConfig, "decision_wait: 30s", "decision_wait: 3s", 1)
```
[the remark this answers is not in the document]
joaopgrassi: I changed it now to `10ms` and it passes still.

testbed/integration/config_examples_test.go
```diff
@@ -0,0 +1,110 @@
+package integration
```
joaopgrassi, line 1: @evan-bradley do you think this package name makes sense? I thought about `samples_e2e` as well, but not sure.

config_examples/README.md
```diff
@@ -0,0 +1,45 @@
+# Collector example configurations
+
+Here you can find a collection of example Collector configurations.
+
+## Samples
+
+- [Drop](drop.yaml)
```
joaopgrassi, line 7: Some of these samples are not clear what they are for/goal. For now I just added the name as the file name, but we should try to improve them in follow ups.

evan-bradley: Agreed, let's tackle that as we implement tests for each.

testbed/integration/config_examples_test.go
evan-bradley: I think this is fine (we're checking that our configs and Collector are compatible), but I don't think the package name is paramount; `samples_e2e` would be fine too. We can always change it later.

testbed/integration/validation.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"fmt"
+	"testing"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/testbed/testbed"
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+)
+
+// SampleConfigsValidator implements TestCaseValidator.
```
evan-bradley, line 12: Just like the other file, asserting the interface implementation here would be nice.

testbed/integration/data_providers.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+var _ testbed.DataProvider = &sampleConfigsDataProvider{}
+
+type sampleConfigsDataProvider struct {
+	traces             ptrace.Traces
+	dataItemsGenerated *atomic.Uint64
+}
+
+// GenerateLogs implements testbed.DataProvider.
```
evan-bradley, line 19: Just a nitpick, but I would say we should remove these comments since we check that it implements the interface above, and that should be sufficient to determine what this method does.

testbed/integration/validation.go
```diff
@@ -0,0 +1,73 @@
...
+		for i := 0; i < rss.Len(); i++ {
+			ss := rss.At(i).ScopeSpans()
+			for j := 0; j < ss.Len(); j++ {
+				spans := ss.At(j).Spans()
+				for k := 0; k < spans.Len(); k++ {
+					recdSpan := spans.At(k)
+					assert.Contains(t,
+						spansMap,
+						traceIDAndSpanIDToString(recdSpan.TraceID(), recdSpan.SpanID()),
+						fmt.Sprintf("Span with ID: '%s' not found among expected spans", recdSpan.SpanID()))
```
evan-bradley, line 51:
```suggestion
						fmt.Sprintf("Span with ID: %q not found among expected spans", recdSpan.SpanID()))
```
Just slightly cleaner, there's a format specifier that handles quoting.

config_examples/README.md
```diff
@@ -0,0 +1,45 @@
...
+so you can also see the data in your Dynatrace environment.
+
+Before running the samples, make sure you have the following environment variables set:
+
+- `DT_OTLP_ENDPOINT`: The OTLP HTTP endpoint of your Dynatrace environment.
+  - Follow the guide: [Export to ActiveGate](https://docs.dynatrace.com/docs/shortlink/otel-getstarted-otlpexport#export-to-activegate)
+    to see how to get the correct API URL for your environment
+- `API_TOKEN`: The Dynatrace API access token. Follow the guide on [Authentication](https://docs.dynatrace.com/docs/shortlink/otel-getstarted-otlpexport#authentication-export-to-activegate) to see the scopes required for ingesting OTLP data.
+
+## Trying out the configuration examples
```
evan-bradley, line 26: Does setting the environment variables in the user's shell pass them through to the Docker container, or do we need to pass `--env` flags?

testbed/integration/data_providers.go
joaopgrassi: Alright I will remove. I mostly based what's in the original `testbed` package.

testbed/integration/validation.go
joaopgrassi: Nice!

config_examples/README.md
joaopgrassi: Yep, forgot about it. I fixed it now and copied/paste the example and traces show up as expected.

evan-bradley (approved):

PR #130, opened 2024-02-09
Add a smoke test that configures all components

evan-bradley:
This doesn't necessarily start every component, but will at least validate that the component exists in the Collector and its configuration will validate. Where possible I've used the default configuration for each component, but if there are required options I tried to take these from the component's readme.

### DIFF
--- a/testbed/smoke/smoke_test.go
+++ b/testbed/smoke/smoke_test.go
@@ -19,32 +19,104 @@ import (
 var execPath = "../../bin/dynatrace-otel-collector"
 
 func TestCollectorStarts(t *testing.T) {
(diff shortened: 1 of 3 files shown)


testbed/smoke/smoke_test.go
```diff
@@ -19,32 +19,104 @@ import (
...
+			col.PrepareConfig(string(cfg))
+
+			err = col.Start(testbed.StartParams{
+				Name:        "dynatrace-otel-collector",
+				LogFilePath: "col.log",
+			})
+			require.NoError(t, err)
+
+			var resp *http.Response
+			require.Eventually(t, func() bool {
```
dyladan, line 91: Didn't know about this. Very cool

dyladan (approved):

PR #134, opened 2024-02-22
Test jaeger example and align examples with docs

pichlermarc:
This PR
- aligns the examples with [our documentation](https://docs.dynatrace.com/docs/extend-dynatrace/opentelemetry/collector/use-cases/jaeger) by
  - using `DT_ENDPOINT` as the env var as opposed to `DT_OTLP_ENDPOINT`
  - using `grpc` as the default for the `jaeger` reciever instead of `thrift_http`
- adds a test for the Jaeger example using the collector testbed 
  - adds a dependency to `github.com/open-telemetry/opentelemetry-collector-contrib/testbed/datasenders` so that we can use the existing Jaeger gRPC data sender.

depends-on: #132 (removing examples that were meant for a different audience)


config_examples/jaeger.yaml
```diff
@@ -1,17 +1,17 @@
...
-  otlphttp:
-    endpoint: https://${env:HOST}/api/v2/otlp
-    headers:
-      Authorization: Api-Token ${env:API_TOKEN}
-
 receivers:
   jaeger:
     protocols:
-      thrift_http:
+      grpc:
```
evan-bradley, line 4: Why gRPC instead of thrift HTTP, is there not a way to test sending data over thrift?

pichlermarc: There is a way but that would mean that we'd have to basically write a thrift http exporter (implementing the `TraceDataSender` interface).

Also, we use `grpc` in our [docs](https://docs.dynatrace.com/docs/extend-dynatrace/opentelemetry/collector/use-cases/jaeger) - so it aligns it with what we have there.

evan-bradley (approved):

PR #139, opened 2024-03-04
test(examples): add and test histogram sum/count-split configuration

pichlermarc:
This PR adds the config to split `sum`/`count` metrics from Histograms from the [Dynatrace Documentation](https://docs.dynatrace.com/docs/extend-dynatrace/opentelemetry/getting-started/metrics/ingest/migration-guide-otlp-exporter#in-collector-additional-features) as well as tests for that config.


testbed/integration/config_examples_test.go
```diff
@@ -202,3 +203,113 @@ func TestConfigJaegerGrpc(t *testing.T) {
...
+	histogramDataPoint.SetCount(3)
+	histogramDataPoint.SetSum(7.5)
+	histogramDataPoint.SetMin(0.5)
+	histogramDataPoint.SetMax(3.5)
+	histogramDataPoint.BucketCounts().Append(1, 0, 0, 0, 2)
+
+	// Split Histogram
+	splitCountMetric := expectedMetrics.AppendEmpty()
+	splitCountMetric.SetName("my.histogram_count")
+	splitCountMetric.SetUnit("custom_unit")
```
pichlermarc, line 259: **Note for reviewers:** does it make sense to have the `unit` be the same for `count` metrics split from `histogram` metrics?

Take [`http.server.request.duration` metric](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-server) for example, it records the duration of requests on the server. The unit for that metric is `s` (seconds), however, the `count` property on that histogram is actually the amount of requests recorded, not seconds (that's the sum, which is the total time the server spent on handling requests of that route, for instance).

pichlermarc: I opened an issue upstream as I think it's applicable to all metrics that can be passed to that function: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/31575

pichlermarc: I suggest we go ahead with adding the tests as-is then iterate on them based on the outcome of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/31575 

WDYT @evan-bradley @arminru?

testbed/integration/metrics_validation.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (v *MetricsSampleConfigsValidator) Validate(tc *testbed.TestCase) {
+	actualMetrics := tc.MockBackend.DataItemsReceived()
+
+	assert.EqualValues(v.t, v.expectedMetrics.MetricCount(), actualMetrics, "Received and expected number of metrics do not match.")
+	assertExpectedMetricsAreInReceived(v.t, []pmetric.Metrics{v.expectedMetrics}, tc.MockBackend.ReceivedMetrics)
+}
+
+func (v *MetricsSampleConfigsValidator) RecordResults(tc *testbed.TestCase) {
+}
```
pirgeo, line 34: Do we need this?

testbed/integration/config_examples_test.go
```diff
@@ -202,3 +203,113 @@ func TestConfigJaegerGrpc(t *testing.T) {
...
+	tc := testbed.NewTestCase(
+		t,
+		dataProvider,
+		sender,
+		receiver,
+		col,
+		validator,
+		&testbed.CorrectnessResults{},
+	)
+	t.Cleanup(tc.Stop)
```
pirgeo, line 295: Is this to make sure nothing else is running in the background?

pirgeo (approved):
Looks good to me. I'll defer to someone with more knowledge of the testbed framework, but from a high level everything looks good!

testbed/integration/config_examples_test.go
pirgeo: Ah, I see. It's like a `defer` for the test. Got it, thank you

PR #150, opened 2024-03-14
Provide example for deriving metrics from pre-sampled traces

evan-bradley:
This configuration provides a basis for using the span metrics connector to obtain request call counts, error counts, and response time durations for traces that arrive at the Collector. These traces can also be sampled, which can be used to reduce total data ingest into Dynatrace.

cc @mviitane @mreider

### DIFF
--- a/config_examples/spanmetrics.yaml
+++ b/config_examples/spanmetrics.yaml
@@ -1,0 +1,71 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
+
+processors:
+  transform:
+    metric_statements:
+    - context: metric
+      statements:
+      # Extract sum and count metrics from the service. We can use these metrics
+      # to get the average request duration by using a metric query like so:
+      # requests.duration_sum:splitBy("service.name")/requests.duration_count:splitBy("service.name")
+      # Note that the resulting values in the above query will be unitless.
+      - extract_count_metric(false) where type == METRIC_DATA_TYPE_HISTOGRAM
+      - extract_sum_metric(false) where type == METRIC_DATA_TYPE_HISTOGRAM
+  filter:
+    metrics:
+      metric:
+      # The Dynatrace OTLP metrics ingest doesn't currently support histograms, so filter them out.
+      - type == METRIC_DATA_TYPE_HISTOGRAM
+  transform/spanmetrics:
+    metric_statements:
+    - context: datapoint
+      statements:
+      # Not required, but simplifies queries in Dynatrace by removing the requirement for a `.count`
+      # suffix on the metric name in the query.
+      - convert_sum_to_gauge() where IsMatch(metric.name, "^requests.*_sum") and metric.type == METRIC_DATA_TYPE_SUM
+    - context: metric
+      statements:
+      # Map the units to something that explicitly counts them in Dynatrace.
+      - set(unit, "{requests}")  where IsMatch(name, "^requests.duration_count")
+      - set(unit, "{requests}")  where IsMatch(name, "^requests.calls")
+  tail_sampling:
+    # This config keeps all traces just as a demonstration. Adjust with policies of your choice.
+    policies:
+      - name: keep-traces
+        type: always_sample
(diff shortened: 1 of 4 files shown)


testbed/integration/config_examples_test.go
```diff
@@ -12,6 +12,7 @@ import (
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/pdata/ptrace"
+	semconv "go.opentelemetry.io/collector/semconv/v1.18.0"
```
arminru, line 15: This file is already getting rather long with a bunch of imports that only apply to certain cases, would it make sense to split the tests into individual files/modules that map 1:1 top the config_example yamls?

evan-bradley: I'm not too concerned about the dependencies here since this is just a test. I don't care either way exactly which files we put the test cases into, but I've stuck with keeping them in the same file for now since we only have a single test case per config example.

config_examples/spanmetrics.yaml
```diff
@@ -0,0 +1,70 @@
...
+    metrics:
+      metric:
+      # The Dynatrace OTLP metrics ingest doesn't currently support histograms, so filter them out.
+      - type == METRIC_DATA_TYPE_HISTOGRAM
+  transform/spanmetrics:
+    metric_statements:
+    - context: datapoint
+      statements:
+      # Not required, but simplifies queries in Dynatrace.
+      - convert_sum_to_gauge() where IsMatch(metric.name, "^requests.*") and metric.type == METRIC_DATA_TYPE_SUM
```
dyladan, line 30: Not sure I understand the reasoning behind this. If gauge is semantically correct that is one thing, but I don't think it is in this case right? Why is the querying easier for gauges than sums? Is this something our users are likely to understand when they see it?

dyladan (commented):
Seems fine but I'm a bit thrown by the converstion from sum to gauge for things that are a count

config_examples/spanmetrics.yaml
evan-bradley: To address each of your questions:
1. No, gauges are not semantically correct here; both of these values can be aggregated.
2. I can't answer that, but converting the `_sum` metric from a sum to a gauge does remove the need for a `.count` suffix.
3. I tried to make the comment as detailed as possible, but I can't say that I fully understand this, so I wouldn't expect our users to understand it either.

config_examples/spanmetrics.yaml
```diff
@@ -0,0 +1,71 @@
...
+      metric:
+      # The Dynatrace OTLP metrics ingest doesn't currently support histograms, so filter them out.
+      - type == METRIC_DATA_TYPE_HISTOGRAM
+  transform/spanmetrics:
+    metric_statements:
+    - context: datapoint
+      statements:
+      # Not required, but simplifies queries in Dynatrace by removing the requirement for a `.count`
+      # suffix on the metric name in the query.
+      - convert_sum_to_gauge() where IsMatch(metric.name, "^requests.*_sum") and metric.type == METRIC_DATA_TYPE_SUM
```
dyladan, line 31: I think I'd still leave this out. Adding a `.count` suffix isn't the end of the world and this confuses the example in my opinion. If sums are difficult to work with that's a product question, not something we should be papering over by ingesting semantically incorrect data.

dyladan (approved):
Looks good. I left a comment but it's not a deal breaker.

config_examples/spanmetrics.yaml
evan-bradley: We should also consider a change to our [histogram conversion snippet](https://github.com/Dynatrace/dynatrace-otel-collector/blob/main/config_examples/split_histogram.yaml#L19) if we want to alter our overall approach here, I'm largely trying to build on and align with that.

I do see your point, I'll remove this for now and we can always add it back in later if we want. My counterpoint to making the metric type semantically incorrect is that this is a workaround for the lack of support for ingesting histograms, so we can bend the rules a little bit. Even if the sum is not an OTLP gauge, it may be semantically close enough to a Dynatrace gauge that we can make that leap in logic.

PR #166, opened 2024-04-17
Include `timetzdata` build flag

joaopgrassi:
Addresses issue where the collector fails to locate timezones when using the Syslog receiver


heckelmann (approved):
LGTM, thank you @joaopgrassi

pichlermarc (approved):

PR #171, opened 2024-04-25
Add k8s enrichment and prometheus e2e tests

joaopgrassi:
This adds e2e tests for the following use-cases

- [Enrich from Kubernetes](https://docs.dynatrace.com/docs/shortlink/otel-collector-cases-k8s-enrich).
- [Scrape data from Prometheus](https://docs.dynatrace.com/docs/shortlink/otel-collector-cases-prometheus)

Since the [k8stest package is internal](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/2dbdca3e2f53d3c4481212a7f147d7199cdeb55f/internal/k8stest), I had to copy the files over. Once that is changed, we can delete the files here. 

It deploys the collector with our recommended transform rules and then assert on the exported traces on the attributes.


testbed/integration/k8senrichment/README.md
```diff
@@ -0,0 +1,17 @@
+# Enrich from Kubernetes
```
joaopgrassi, line 1: I don't love the directory structure here. It would be nice if all the utilities could maybe stay in testbed or even better, in internal, and then we can have a `tests/usecases` folder where we can place all the usecase tests we want to test.

testbed/integration/k8senrichment/e2e_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	for _, tc := range tcs {
+		scanTracesForAttributes(t, tracesConsumer, tc.service, tc.attrs)
+	}
+}
+
+func scanTracesForAttributes(t *testing.T, ts *consumertest.TracesSink, expectedService string,
+	kvs map[string]*expectedValue) {
+	// Iterate over the received set of traces starting from the most recent entries due to a bug in the processor:
+	// https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18892
```
evan-bradley, line 166: I assume this was copied from upstream, but the issue has been resolved. Did the associated PR not fix this problem?

testbed/integration/prometheus/e2e_test.go
```diff
@@ -0,0 +1,144 @@
...
+	testDir := filepath.Join("testdata", "prometheus")
+	nsFile := filepath.Join(testDir, "install.sh")
+	cmd, err := exec.Command("/bin/bash", nsFile).Output()
+
+	if err != nil {
+		return fmt.Errorf("Failed to install Prometheus node exporter Helm chart %s", err)
+	}
+
+	fmt.Print(string(cmd))
+
```
evan-bradley, line 86:
```suggestion
```
We should probably remove this, right?

testbed/integration/prometheus/e2e_test.go
```diff
@@ -0,0 +1,144 @@
...
+	}()
+
+	wantEntries := 2 // Minimal number of metric requests to wait for.
+	waitForData(t, wantEntries, metricsConsumer)
+
+	expectedMetrics := []string{
+		"otelcol_process_memory_rss", "scrape_duration_seconds", "scrape_samples_post_metric_relabeling",
+	}
+
+	scanForServiceMetrics(t, metricsConsumer, "opentelemetry-collector", expectedMetrics)
```
evan-bradley, line 73: Should we also check for metrics from the node exporter?

evan-bradley (commented):
Overall I think this looks good, just a few notes.

testbed/integration/k8senrichment/README.md
joaopgrassi: Alright, I will keep it for a separated PR and try to improve the structure.

testbed/integration/prometheus/e2e_test.go
joaopgrassi: It is helpful because it prints the results from the help commands, which show when the Prometheus node exporter is ready. So I think keeping it makes sense.

testbed/integration/k8senrichment/e2e_test.go
joaopgrassi: Ah good find, I meant to remove this after the clean up but forgot. I will remove!

testbed/integration/prometheus/e2e_test.go
joaopgrassi: Makes sense! I added two metrics to check from the node exporter now.

testbed/integration/prometheus/e2e_test.go
evan-bradley: That's fair. This is minor, but could you add a comment above this line repeating what you just described? Raw print statements stick out, so that way we know why this is here in the future.

evan-bradley (approved):

PR #175, opened 2024-05-01
Add a confmap.Provider that communicates with the EEC

evan-bradley:
This forks the upstream HTTP(s) Provider and adds configuration options exposed through query parameters inside the URL fragment that is given to the provider.


confmap/provider/eecprovider/provider_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"go.opentelemetry.io/collector/confmap/confmaptest"
+)
+
+func TestSupportedScheme(t *testing.T) {
+	fp := NewFactory().Create(confmaptest.NewNopProviderSettings())
+	assert.Equal(t, "http", fp.Scheme())
```
heckelmann, line 18: I'm not an expert, but shouldn't this be `eec`?

evan-bradley: Good catch, it should be. I've updated this.

confmap/provider/internal/configurablehttpprovider/provider.go
```diff
@@ -0,0 +1,194 @@
...
+	"go.opentelemetry.io/collector/confmap"
+)
+
+type SchemeType string
+
+const (
+	// "EEC Insecure" - equivalent to HTTP
+	EECIScheme SchemeType = "eeci"
+	// equivalent to HTTPS
+	EECScheme SchemeType = "eec"
```
evan-bradley, line 28: I went with a "secure by default" approach here with having the extra letter indicate an insecure connection instead of being required to use TLS.

I realize that this goes against the "HTTP"/"HTTPS" convention, so if we don't think this is worth it, I can update this to be something like "eec"/"eecs".

heckelmann: To be honest, I was confused the first time I saw the `eeci` SchemeType, as I assumed it should be `eec` / `eecs` cause your given reason.

I don't have a strong opinion on that but would be slightly in favor of `eecs` for old grumpy admins like I am ;)

.github/workflows/e2e.yaml
```diff
@@ -40,6 +40,12 @@ jobs:
           echo "Building collector artifact"
           make build
 
+      - name: Upload binary artifact
```
dyladan, line 43: Did you mean to make these changes in the same PR?

confmap/provider/eecprovider/README.md
```diff
@@ -0,0 +1,28 @@
+# EEC Provider
+
+This is an OpenTelemetry Collector `confmap.Provider` module that allows the
+Collector to be configured with the Dynatrace Extensions Execution Controller.
+
+> [!WARNING]
+> This is not intended for customer use, but is only intended for use by
+> Dynatrace when managing its own Collector deployments. As such, issues with
+> this component will fall outside our support policy.
```
dyladan, line 9: I think "only intended for use by Dynatrace when managing its own Collector deployments" makes it sound like it's only for use within Dynatrace the SaaS cluster or something like that. It's not clear that it's for use by Dynatrace in customer environments.

```suggestion
> This is not intended for direct customer use, but is only intended for use by
> Collector installations installed and managed by Dynatrace. Configuring this
> component directly is not supported.
```

confmap/provider/eecprovider/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Options
+
+| Key | Default | Description |
+|-----|---------|-------------|
+| auth-env |  None | An environment variable that should be read to get a plaintext API token or other key to be used in an HTTP header. Mutually exclusive from `auth-file`, passing both options should result in an error. |
+| auth-file | None | A filepath containing a plaintext version of an API token or other key to be used in an HTTP header to authenticate with the EEC. |
+| auth-header | Api-Token | A header that will be used to authenticate the provider with the EEC. |
+| refresh-interval | 10s | A time duration that defines how frequently the provider should check the given URL for updates. |
+| insecure | false | If set to "true", use HTTP for the connection to the server. If unset or set to "false", use HTTPS. |
```
dyladan, line 28: I think we should leave use out of public documentation if we're going to discourage direct use? We can either document internally or rely on the code comments in the config struct

confmap/provider/eecprovider/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+type config struct {
+	// A time duration that defines how frequently the provider should check the given URL for updates.
+	refreshInterval time.Duration
+
+	// A header that will be used to authenticate the provider with the EEC.
+	// We will always use Api-Token (and this can be assumed as a default if auth-file is passed),
+	// but upstream will want to support configurable headers and we should consider
+	// offering this option should upstream determine the option is required.
+	authHeader string
```
dyladan, line 22: I'd leave this out since it would also likely need configurable format for upstream. For our internal use we already know there is only 1 valid value.

confmap/provider/eecprovider/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// Token passed as the value for the header in authHeader.
+	authToken string
+
+	// Whether to use HTTP instead of HTTPS
+	insecure bool
```
dyladan, line 28: Can we infer this from url scheme like `eec://` and `eecs://`? Maybe HTTP is getting in the way of my brain but it strikes me as a bit odd to include it as a query parameter.

confmap/provider/eecprovider/provider.go
```diff
@@ -0,0 +1,199 @@
...
+		cert, err := os.ReadFile(filepath.Clean(p.caCertPath))
+
+		if err != nil {
+			return nil, fmt.Errorf("unable to read CA from %q URI: %w", p.caCertPath, err)
+		}
+
+		if ok := pool.AppendCertsFromPEM(cert); !ok {
+			return nil, fmt.Errorf("unable to add CA from uri: %s into the cert pool", p.caCertPath)
+		}
+	}
```
dyladan, line 87: This is all only required for testing right?

confmap/provider/eecprovider/watcher.go
```diff
@@ -0,0 +1,51 @@
...
+	configHash      [32]byte
+}
+
+func (w *watcher) watchForChanges() {
+	ticker := time.NewTicker(w.refreshInterval)
+
+	for {
+		select {
+		case <-ticker.C:
+			body, err := w.getConfigBytes()
```
dyladan, line 32: What happens if this takes longer than `refreshInterval`? Should we terminate the HTTP request if that happens?

manifest.yaml
```diff
@@ -39,3 +39,14 @@ processors:
...
+providers:
+  - gomod: go.opentelemetry.io/collector/confmap/provider/envprovider v0.99.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/fileprovider v0.99.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpprovider v0.99.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpsprovider v0.99.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/yamlprovider v0.99.0
+  - gomod: github.com/Dynatrace/dynatrace-otel-collector/confmap/provider/eecprovider v0.4.0
+
+replaces:
+  - github.com/Dynatrace/dynatrace-otel-collector/confmap/provider/eecprovider => ../confmap/provider/eecprovider
```
dyladan, line 52: I assume this will not be required after this PR is merged to main?

dyladan (commented):
Looks generally good just have a few comments

.github/workflows/e2e.yaml
evan-bradley: This is step necessary to grab the binary later on when we run the E2E tests for the provider. The rest of the changes are just to run the tests.

confmap/provider/eecprovider/README.md
evan-bradley: Yeah I wasn't really satisfied with my wording. I like yours better, thanks.

.github/workflows/e2e.yaml
dyladan: update: i see it's required later in e2e tests

confmap/provider/eecprovider/README.md
evan-bradley: I'm fine with whatever, this just seemed like the ideal place to have config and the code comments may not be easy to locate. I think any sufficiently-motivated user could figure it out, so I was just going to rely on our disclaimer to discourage users. That said, I can strip down this file if we think it's a risk.

confmap/provider/eecprovider/config.go
evan-bradley: I suspect this option will be for local testing or edge cases, which is why I "hid" it in an option rather than offered it as a supported scheme. We can use the scheme, but I think for our specific use case it's a worse option.

manifest.yaml
evan-bradley: I think it will be always required. We release the distro and module versions at the exact same time, so the tag for the module won't exist when we go to build the distro.

confmap/provider/eecprovider/provider.go
evan-bradley: Right. there's no way to set `caCertPath` outside of this package.

confmap/provider/eecprovider/watcher.go
evan-bradley: That's a really good observation, I overlooked that. I've updated it so we're properly propagating context down to the request.

confmap/provider/eecprovider/config.go
evan-bradley: That's fine with me, done.

dyladan (approved):

confmap/provider/eecprovider/README.md
```diff
@@ -0,0 +1,28 @@
...
+```text
+dynatrace-otel-collector --config=eec://my.eec.host:31098#refresh-interval=5s&auth-header=Api-Token&auth-file=/var/private/token.key
+```
+
+### Options
+
+| Key | Default | Description |
+|-----|---------|-------------|
+| auth-env |  None | An environment variable that should be read to get a plaintext API token or other key to be used in an HTTP header. Mutually exclusive from `auth-file`, passing both options should result in an error. |
+| auth-file | None | A filepath containing a plaintext version of an API token or other key to be used in an HTTP header to authenticate with the EEC. |
```
odubajDT, line 25: `auth-header` is present in the example, but missing here, should we add it?

confmap/provider/eecprovider/README.md
```diff
@@ -0,0 +1,28 @@
...
+> This is not intended for direct customer use, but is only intended for use by
+> Collectors installed and managed by Dynatrace. Configuring this
+> component directly is not supported.
+
+## Configuration
+
+This confmap Provider can be configured by pasing query parameter-formatted values
+inside the fragment of the URL given to the config flag. For example:
+
+```text
```
odubajDT, line 16:
```suggestion
```shell
```

internal/mockeec/README.md
```diff
@@ -0,0 +1,14 @@
+# Mock EEC server
+
+## How to use
+
+1. Start the server in this directory: `go run .`
+2. Run the Collector:
+
+   ```text
+    ./bin/dynatrace-otel-collector --config=eec://localhost:8000/config.yaml#insecure=true&refresh-interval=100ms
+   ```
```
odubajDT, line 10:
```suggestion
```shell
./bin/dynatrace-otel-collector --config=eec://localhost:8000/config.yaml#insecure=true&refresh-interval=100ms
```
```

File: .github/workflows/e2e.yaml [hunk 14]
```diff
@@ -128,3 +134,41 @@ jobs:
...
+        with:
+          name: dynatrace-otel-collector-bin
+          path: bin
+
+      - run: chmod +x bin/*
+
+      - name: Run e2e tests
+        run: |
+          cd confmap/provider/eecprovider
+          go test -v --tags=e2e
```
odubajDT, line 174: are unit tests ran as part of this as well? Or are they part of a different workflow?

internal/mockeec/main.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"flag"
+	"log"
+	"net/http"
+)
+
+func main() {
+	port := flag.String("p", "8000", "port to listen on")
+	dir := flag.String("d", ".", "directory to serve files from")
```
odubajDT, line 11: does it make sense to make the mockeec at least a little bit configurable? Maybe for future usage in a different component

confmap/provider/eecprovider/e2e_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetsConfig(t *testing.T) {
+	f, err := os.ReadFile("./testdata/otel-config.yaml")
+	require.NoError(t, err)
+	fs := &fileserver{config: f}
+	ts := httptest.NewServer(http.HandlerFunc(fs.HandleRequest))
+	cmd := exec.Command("../../../bin/dynatrace-otel-collector", fmt.Sprintf("--config=%s", configureProvider(t, ts.URL)))
```
odubajDT, line 23: would it makes sense to mock the usage to the `dynatrace-otel-collector`? This way a bug in a different component can cause test failure here. I understand this is an e2e test, but maybe I would prefer to do e2e testing on a real k8s cluster (including manifests etc). WDYT?

confmap/provider/eecprovider/yaml.go
```diff
@@ -0,0 +1,21 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package eecprovider // import "github.com/Dynatrace/dynatrace-otel-collector/confmap/provider/eecprovider"
```
odubajDT, line 4: I would maybe move this to a separate package, since it can be used also outside of this provider

manifest.yaml
```diff
@@ -39,3 +39,14 @@ processors:
...
   - gomod: go.opentelemetry.io/collector/connector/forwardconnector v0.99.0
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/connector/spanmetricsconnector v0.99.0
+
+providers:
+  - gomod: go.opentelemetry.io/collector/confmap/provider/envprovider v0.99.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/fileprovider v0.99.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpprovider v0.99.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpsprovider v0.99.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/yamlprovider v0.99.0
+  - gomod: github.com/Dynatrace/dynatrace-otel-collector/confmap/provider/eecprovider v0.4.0
```
joaopgrassi, line 49: What about calling this `v0.0.0` since there's not an actual version involved here? 🤔

confmap/provider/eecprovider/README.md
```diff
@@ -0,0 +1,28 @@
...
+dynatrace-otel-collector --config=eec://my.eec.host:31098#refresh-interval=5s&auth-header=Api-Token&auth-file=/var/private/token.key
+```
+
+### Options
+
+| Key | Default | Description |
+|-----|---------|-------------|
+| auth-env |  None | An environment variable that should be read to get a plaintext API token or other key to be used in an HTTP header. Mutually exclusive from `auth-file`, passing both options should result in an error. |
```
joaopgrassi, line 24: "should result in an error" or will result in an error? 🤔

confmap/provider/eecprovider/README.md
```diff
@@ -0,0 +1,28 @@
+# EEC Provider
+
+This is an OpenTelemetry Collector `confmap.Provider` module that allows the
+Collector to be configured with the Dynatrace Extensions Execution Controller.
+
+> [!WARNING]
+> This is not intended for direct customer use, but is only intended for use by
```
arminru, line 7:
```suggestion
> This is an internal component not intended for direct customer use, but is only intended for use by
```

confmap/provider/eecprovider/README.md
evan-bradley: I actually meant to get rid of `auth-header`. I'll remove it from the example.

confmap/provider/eecprovider/README.md
evan-bradley: I tried this, but the problem I ran into is that the config fragments are highlighted as comments, at least in my local Markdown renderer. I'll give it a shot with the GitHub renderer and see if it's any different.

.github/workflows/e2e.yaml
evan-bradley: Unit tests are run as part of our main CI workflow, specifically at this step: https://github.com/Dynatrace/dynatrace-otel-collector/blob/main/.github/workflows/ci.yaml#L43.

internal/mockeec/main.go
evan-bradley: What other options do you think would be helpful? My goal here is essentially just to have this be a basic fileserver for quick demonstration with a user-run Collector, particularly to see reloads work in action.

confmap/provider/eecprovider/e2e_test.go
evan-bradley: >would it makes sense to mock the usage to the `dynatrace-otel-collector`?

We could directly utilize the libraries the Collector uses to instantiate this component and test it that way, but I think this should be fairly reliable and saves us some effort. It also tests the e2e flow of building a Collector with this component (manifest.yaml -> OpenTelemetry Collector Builder -> Binary -> ...), which isn't strictly in the scope of these tests, but for the way we do "E2E" tests (tests that work with a built binary) I think it's okay. It will also alert us if any upstream changes break us in this process.

>This way a bug in a different component can cause test failure here.

For the most part we're only testing the pieces we are concerned about here, which are the modules that manage the Collector's configuration and internal lifecycle. We do depend a little bit on other components not to panic on startup, or on any extensions to actually respond to an HTTP request, but I don't expect that functionality will be fragile.

>I see these tests more as component tests than an e2e test. To do a proper e2e testing maybe I would prefer to do it on a real k8s cluster (including manifests etc). WDYT?

To me that would depend on what you're considering the system under test. In this case, I would say that the component is the confmap Provider and the system is the whole Collector + a configuration server, so this is E2E from the sense that we are testing the interaction between the Collector and configuration server with a full live Collector. I didn't want to bring any infrastructure setups into these tests since the functionality should behave the same regardless of whether we run in a VM, k8s, FaaS, etc.

That said, it's good to make sure we're covering all our bases. Given the above, do you think we're still missing something from these tests by not testing a Collector deployed to k8s?

confmap/provider/eecprovider/provider.go
```diff
@@ -0,0 +1,208 @@
...
+
+const (
+	EECScheme SchemeType = "eec"
+)
+
+const ApiTokenHeader = "Api-Token"
+
+type provider struct {
+	caCertPath         string // Used for tests
+	insecureSkipVerify bool   // Used for tests
```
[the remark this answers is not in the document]
evan-bradley: In general I would agree. These were taken from the upstream code, and I didn't want to do a major refactor when I hope that we should be able to use the upstream Provider in the future. I think the lack of a really solid way to mock in Go also contributes to this being more of a valid approach, though I haven't explored alternatives for this case in particular.

confmap/provider/eecprovider/yaml.go
evan-bradley: That's a good callout, but I would prefer to wait until we need it to be used elsewhere. Right now we only plan to have a single Provider, and we also hope that we will be able to remove this Provider in the future in favor of using the upstream HTTP/HTTPS providers.

manifest.yaml
evan-bradley: That's a fair point considering the replace statement. Done.

confmap/provider/eecprovider/README.md
evan-bradley: Good catch. "Will result in an error" is right.

confmap/provider/eecprovider/README.md
evan-bradley: You can see what it looks like here: https://github.com/Dynatrace/dynatrace-otel-collector/commit/eec2b66eb3dd4d3c8bc3b62853a88decebed54ad?short_path=5b27679#diff-5b27679011dbb54249c6f87e02849959e30682d84058764b1a8bedfd3957c6c7

I think the highlighted ampersand is more distracting than anything, I'd like to opt to leave it as plain text for now. Let me know if you think the highlighting would still be worth it.

odubajDT: it's just a nitpick. I leave the decision on you.

internal/mockeec/main.go
odubajDT: For example reading the port and directory from command line parameters, nothing serious here. It can cause problems for somebody running it locally, since port 8000 can be already in use.

heckelmann: This is a command line parameter and not a static value.

odubajDT: ah sorry I misread the code. Ignore my comments please

joaopgrassi (approved):

heckelmann (approved):
LGTM

confmap/provider/eecprovider/e2e_test.go
odubajDT: If the rest of the team is ok with this, I agree to leave it as it is

confmap/provider/eecprovider/yaml.go
odubajDT: It's a note for the future, I'll leave it up to you

confmap/provider/eecprovider/e2e_test.go
odubajDT: In general using binaries in tests is not the best approach, although I agree it's the easier path. Still a proper E2E testing should happen on a cluster/container (since that's the expected env where the provider will run on), as with this approach we are not able to mock the environment behavior (such as network lag etc.) It should basically mock the user (machine) behavior. 

In my eyes we might want to re-evaluate these to component tests -> those should check the behavior of the component (provider) and prove that the component is responding to the system on different requests as expected, no matter from when the requests come from. Therefore we can mock the behavior of the used binary. Also it may be problematic to run them locally, since the test requires a pre-requisite of building the binary locally and moving it to the correct folder. Tests in general should be completely independent from other dependencies (except e2e). For example, missing of the binary will negatively affect the result of the test.

The decision is completely on you, I wanted to just point out the things that can bite us in the future.

confmap/provider/eecprovider/provider.go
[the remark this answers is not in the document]
evan-bradley: >What mocking options do you miss in go?

Mostly the ability to override concrete types. Interfaces would be able to achieve this.

I also agree that we shouldn't have anything testing-specific in production code if possible. If we decide to keep this Provider longer-term instead of switching to the upstream Providers, we can revisit this and refactor it to our standards. Right now I expect it will be removed within a few months.

confmap/provider/eecprovider/e2e_test.go
evan-bradley: I think we're largely on the same page here, though I still think that deploying the Collector binary to test this code is too much setup for what essentially amounts to testing an HTTP request. If we were testing the whole Collector (env vars, resource limits, reading files, networking, etc.) I would be more inclined to suggest the same.

Right now the tests are gated behind a [e2e build flag](https://github.com/Dynatrace/dynatrace-otel-collector/pull/175/files#diff-873ab01c873087589fdce6adcee1d3dad3c77be37850bc0977a76c47dbcc5f93R1), so they will only be run if the flag is enabled, which we explicitly [do in the workflow](https://github.com/Dynatrace/dynatrace-otel-collector/pull/175/files#diff-a2de7550554c0198ac7f959c78a19ee0996921c98034411de47a7dd49b0c209bR174). Running `go test` locally won't have any dependencies on the binary unless whoever is running the tests also enables the flag.

>The decision is completely on you, I wanted to just point out the things that can bite us in the future.

Thanks for giving it some serious consideration. I'm trying to keep things simple for now because I expect this code will be temporary and seldom-modified while we have discussions in the OTel community to determine how to bring this functionality to the upstream Providers. If we end up needing to keep it long-term we can return to this and see what we can do to ensure our development approach is more robust.

PR #184, opened 2024-05-10
Add arg to read config from default location

evan-bradley:


dyladan (approved):

dyladan (changes requested):
This still doesn't work. The docs point to `/etc/dynatrace-otel-collector/config.yaml`

dyladan (approved):
I misremembered. This is correct https://docs.dynatrace.com/docs/extend-dynatrace/opentelemetry/collector/configuration

PR #188, opened 2024-05-14
Add section to list which container registries are available to pull images from

joaopgrassi:


README.md
```diff
@@ -15,6 +15,13 @@ For configuration examples, please see [`config_examples`].
 [Dynatrace's documentation]: https://docs.dynatrace.com/docs/extend-dynatrace/opentelemetry/collector/deployment
 [`config_examples`]: ./config_examples/README.md
 
+### Container images
+
+Container images for the Dynatrace distribution of the [OpenTelemetry Collector] are available in:
```
evan-bradley, line 20:
```suggestion
Container images for the Dynatrace distribution of the OpenTelemetry Collector are available in:
```
I don't think we need to link this again since we already did above.

evan-bradley (approved):

PR #198, opened 2024-05-24
Prepare v0.6.0 changelog

joaopgrassi:
Since this is the first release with the new changelog process, we will have to manually add the things we did as there's no changelog entry files. We can grab those from the automated release notes GH generates once we create the tag and release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,12 +19,12 @@ https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v
 <summary>Highlights from the upstream Collector changelog</summary>
 </br>
 
-🛑 Breaking changes 🛑
+### 🛑 Breaking changes 🛑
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,3 +3,31 @@
 # Changelog
 
 <!-- next version -->
+
+## v0.6.1
```
pirgeo, line 7: Why is this 0.6.1 and not 0.6.0?

pirgeo (commented):
So, if I understand correctly, the release process works by merging this PR and then creating a release tag for it at 0.6.0, which is then the new release (and triggers pushing the binaries)?

evan-bradley:
>So, if I understand correctly, the release process works by merging this PR and then creating a release tag for it at 0.6.0, which is then the new release (and triggers pushing the binaries)?

That's right, the goal of this PR is to take any final steps before cutting the new release. For us that's only generating the changelog.

PR #211, opened 2024-06-11
test(examples/syslog): add test and example for syslog

pichlermarc:
This PR adds the syslog ingest example from the [Dynatrace Documentation](https://docs.dynatrace.com/docs/extend-dynatrace/opentelemetry/collector/use-cases/syslog) as well as integration tests for it.

### DIFF
--- a/.chloggen/feat_int-test-syslog.yaml
+++ b/.chloggen/feat_int-test-syslog.yaml
@@ -1,0 +1,13 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
+component: 'examples'
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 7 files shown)


joaopgrassi (commented):
The changes and test looks good to me, but I wonder if it wouldn't be nicer if we move the test to its own folder, like we did for prometheus and k8s? Then you could also add a README file like I did here https://github.com/Dynatrace/dynatrace-otel-collector/tree/main/internal/testbed/integration/prometheus

pichlermarc:
> The changes and test looks good to me, but I wonder if it wouldn't be nicer if we move the test to its own folder, like we did for prometheus and k8s? Then you could also add a README file like I did here https://github.com/Dynatrace/dynatrace-otel-collector/tree/main/internal/testbed/integration/prometheus

Hmm, should we actually split up all of them then (jaeger, syslog, histogram-split, and otlp)? :thinking: 
If we'd decide to do so I'd do that in a follow-up as a pure refactoring PR to keep the diff manageable.

joaopgrassi (approved):
>Hmm, should we actually split up all of them then (jaeger, syslog, histogram-split, and otlp)? 🤔
If we'd decide to do so I'd do that in a follow-up as a pure refactoring PR to keep the diff manageable.

Sounds good 👍

PR #222, opened 2024-06-25
[chore] document local e2e test execution, minor tweaks to tests

bacherfl:


internal/testbed/integration/k8senrichment/README.md
```diff
@@ -15,3 +15,80 @@ on the k8s cluster with configurations as per the Dynatrace documentation page.
...
+```shell
+kind create cluster
+```
+### Build the collector binary
+
+This is done using the 
+```shell
+make build
+```
+command. This will build the collector distro, and place the built binary
```
mowies, line 36: I suggest putting this inline if it's inside a sentence.
```suggestion
This is done using the `make build` command.
This will build the collector distro, and place the built binary 
```

internal/testbed/integration/k8senrichment/README.md
```diff
@@ -15,3 +15,77 @@ on the k8s cluster with configurations as per the Dynatrace documentation page.
...
+### Build the collector binary
+
+This is done using the `make build` command.
+This will build the collector distro, and place the built binary 
+into the `bin` directory of your local copy of the repository.
+
+**NOTE:** When using an M1 mac, the `make build` command will build the `arm64` binary, which
+will not be able to run as a container on a `kind` cluster.
+For now, as workaround you will need to use the `make build-all` command to build all binaries and then copy the
+`linux_amd64` binary from the `dist` directory into `bin` under the name `dynatrace-otel-collector`.
```
mowies, line 39: why is the ARM version not working? shouldn't that be used for ARM based Macs?

internal/testbed/integration/k8senrichment/testdata/collector/deployment.yaml
```diff
@@ -21,7 +21,7 @@ spec:
           command:
             - /dynatrace-otel-collector
             - --config=/conf/relay.yaml
-          image: "dynatrace-otel-collector:e2e-test"
+          image: "{{ .ContainerRegistry }}dynatrace-otel-collector:e2e-test"
```
odubajDT, line 24: this should be done also for prometheus deployment

internal/testbed/integration/k8senrichment/README.md
bacherfl: I do not know - I just found that for now this is what I had to do to make this run in my setup. Investigating this further could be done in a follow up to make the local setup a bit easier though

internal/testbed/integration/k8senrichment/README.md
```diff
@@ -15,3 +15,77 @@ on the k8s cluster with configurations as per the Dynatrace documentation page.
...
+kube config located in `~/.kube.config`).
+Also, if you are using `podman`, the collector image will be prefixed with `localhost/` within the local
+`kind` registry. In this case, you will need to set the `CONTAINER_REGISTRY` to `localhost/`.
+When using `docker`, setting the `CONTAINER_REGISTRY` env var is not required.
+Below are the commands to execute the `k8senrichment` e2e test:
+
+** Using podman:**
+```shell
+cd internal/testbed/integration/k8senrichment
+KUBECONFIG="/Users/my-user/.kube/config" CONTAINER_REGISTRY="localhost/" go test -v --tags=e2e
```
mowies, line 82: I think this would make more sense. At least, it's easier to copy/paste :)
```suggestion
KUBECONFIG="~/.kube/config" CONTAINER_REGISTRY="localhost/" go test -v --tags=e2e
```

internal/k8stest/k8s_collector.go
```diff
@@ -20,7 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-func CreateCollectorObjects(t *testing.T, client *K8sClient, testID string, manifestsDir string) []*unstructured.Unstructured {
+func CreateCollectorObjects(t *testing.T, client *K8sClient, testID string, manifestsDir string, containerRegistry string) []*unstructured.Unstructured {
```
joaopgrassi, line 23: Instead of the new param here for the container registry, couldn't we do the same thing as in `client.go` where we try to read the env variable here? Then we would save from the calling code to pass it.

joaopgrassi (approved):
I tried the setup with Docker and worked for me. Just one minor comment on the new container registry parameter.

internal/k8stest/k8s_collector.go
bacherfl: yes good point - will adapt that

PR #225, opened 2024-07-02
feat: add zipkin receiver

bacherfl:
Note: Right now there are some duplications in the tests. Since this affects also the already present tests I would suggest that I do some refactoring in a follow up PR, after this one and https://github.com/Dynatrace/dynatrace-otel-collector/pull/222 have been merged


joaopgrassi (approved):

internal/testbed/integration/zipkin/e2e_test.go
```diff
@@ -0,0 +1,226 @@
...
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/component/componenttest"
+	"go.opentelemetry.io/collector/consumer/consumertest"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/receiver/otlpreceiver"
+	"go.opentelemetry.io/collector/receiver/receivertest"
+	"go.uber.org/multierr"
```
evan-bradley, line 22: Nit: we can use `errors.Join` instead of `multierr` now.

evan-bradley (approved):

internal/testbed/integration/zipkin/e2e_test.go
bacherfl: good suggestion, thanks - Since I have also a [refactoring PR](https://github.com/Dynatrace/dynatrace-otel-collector/pull/233) for the integration tests open right now, I will incorporate that in the refactoring PR, as this is one of the functions that have been changed/moved there

PR #227, opened 2024-07-03
[chore] update bind addresses in config examples to localhost

mowies:


config_examples/jaeger.yaml
```diff
@@ -2,7 +2,7 @@ receivers:
   jaeger:
     protocols:
       grpc:
-        endpoint: 0.0.0.0:14250
+        endpoint: localhost:14250
```
joaopgrassi, line 5: Do we want to change to localhost? (which is the default now) or keep as `0.0.0.0` which is likely to be the work that will work OOTB for users? I feel we should leave `0.0.0.0` here still. Maybe we just leave a comment saying this is not to be used in production and etc?

PR #229, opened 2024-07-03
Add statsd receiver and e2e tests

dyladan:
## Changes

- added `statsdreceiver` into the collector
- introduced default configuration for `statsdreceiver`
- e2e tests using the default configuration


internal/testbed/integration/statsd/e2e_test.go
```diff
@@ -0,0 +1,136 @@
...
+				return fmt.Errorf("Expected metric value %f, received %f", expectedVal, datapoint.DoubleValue())
+			}
+			val, ok := datapoint.Attributes().Get(expectedAttrKey)
+			if !ok {
+				return fmt.Errorf("Expected metric attribute not found")
+			}
+			if val.Str() != expectedAttrVal {
+				return fmt.Errorf("Expected metric attribute value %s not found, got %s", expectedAttrVal, val.Str())
+			}
+			return nil
```
bacherfl, line 110: i think this needs to be removed - otherwise the remaining expected metrics will not be checked

internal/testbed/integration/statsd/e2e_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	t.Fatalf("no metric found")
+}
+
+func assertExpectedMetrics(sm pmetric.MetricSlice) error {
+	expectedName := "test.metric"
+	expectedVal := 42.0
+	expectedAttrKey := "myKey"
+	expectedAttrVal := "myVal"
+	expectedTimerName := "test.metric"
```
bacherfl, line 90:
```suggestion
	expectedTimerName := "timerMetric"
```

internal/testbed/integration/statsd/e2e_test.go
```diff
@@ -0,0 +1,136 @@
...
+				return fmt.Errorf("Expected timer metric sum %f, received %f", expectedTimerSum, datapoint.Sum())
+			}
+			val, ok := datapoint.Attributes().Get(expectedTimerAttrKey)
+			if !ok {
+				return fmt.Errorf("Expected timer metric attribute not found")
+			}
+			if val.Str() != expectedTimerAttrVal {
+				return fmt.Errorf("Expected timer metric attribute value %s not found, got %s", expectedTimerAttrVal, val.Str())
+			}
+			return nil
```
bacherfl, line 132: i think this needs to be removed - otherwise the remaining expected metrics will not be checked

internal/testbed/integration/statsd/e2e_test.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+			val, ok := datapoint.Attributes().Get(expectedTimerAttrKey)
+			if !ok {
+				return fmt.Errorf("Expected timer metric attribute not found")
+			}
+			if val.Str() != expectedTimerAttrVal {
+				return fmt.Errorf("Expected timer metric attribute value %s not found, got %s", expectedTimerAttrVal, val.Str())
+			}
+		} else {
+			return fmt.Errorf("Expected metrics not found")
```
odubajDT, line 131:
```suggestion
			return fmt.Errorf("Unexpected metric received: %s", sm.At(i).Name())
```

internal/testbed/integration/statsd/e2e_test.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func scanForServiceMetrics(t *testing.T, ms *consumertest.MetricsSink) {
+	for _, r := range ms.AllMetrics() {
+		for i := 0; i < r.ResourceMetrics().Len(); i++ {
+			sm := r.ResourceMetrics().At(i).ScopeMetrics().At(0).Metrics()
+			assert.NoError(t, assertExpectedMetrics(sm))
+		}
+	}
+	return
```
odubajDT, line 81:
```suggestion
```

internal/testbed/integration/statsd/e2e_test.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+const (
+	equal = iota
+	regex
+	exist
+	testKubeConfig   = "/tmp/kube-config-collector-e2e-testing"
+	kubeConfigEnvVar = "KUBECONFIG"
+	uidRe            = "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
+)
```
bacherfl, line 26: These seem to not be used anymore
```suggestion

```

bacherfl (approved):

internal/testbed/integration/statsd/testdata/statsd/job.yaml
```diff
@@ -0,0 +1,19 @@
...
+  template:
+    spec:
+      containers:
+      - name: netcat
+        image: alpine
+        command: ["/bin/sh", "-c"]
+        args: [
+          "echo 'test.metric:42|g|#myKey:myVal' | nc -u -w 1 -v otelcolsvc.e2estatsd.svc.cluster.local 8125 && \
+           sleep 1 && \
+           echo 'timerMetric:320|ms|@0.1|#timerKey:timerVal' | nc -u -w 1 -v otelcolsvc.e2estatsd.svc.cluster.local 8125"
```
joaopgrassi, line 16: Sorry I'm not familiar with StatsD but I just wanted to make sure: does this metric ends up "touching"/testing the processor config where we convert it to a histogram, prior to exporting it as a gauge?

dyladan: yes it does. the `ms` means it is a timer. There are 3 types of statsd metrics which the config converts to histograms and timer is one of them. This particular entry means the value of `320` should be entered into the histogram 10 times (`@0.1` means it is sampled at a rate of 10%).

dyladan: >prior to exporting it as a gauge
it should not be exported as a gauge. it should be converted to histogram and that's it

dyladan (commented):
Looks good to me. I can't approve as I am the PR author 😆

It looks like you're only testing gauge and timer. Probably fine since all histogram-aggregated metrics should be treated the same anyway, but maybe we should also test counters.

We probably should have split this to 2 PRs, but the history should still accurately reflect credit I think.

odubajDT:
> Looks good to me. I can't approve as I am the PR author 😆
> 
> It looks like you're only testing gauge and timer. Probably fine since all histogram-aggregated metrics should be treated the same anyway, but maybe we should also test counters.
> 

Agree, but this is e2e testing which means not all use-cases need to be tested here. In general e2e tests should in my opinion confirm that the module in general is able to cooperate within a wider context (with processors, exporters etc.). All possible use-cases should have been covered by unit tests in upstream -> we should always aim for fast lightweight unit tests to cover most of the functionality instead of testing the component for many scenarios on a real cluster. I will try to double check the status of tests in upstream and enhance it rather directly there, so we move the maintenance of the tests to the community. WDYT?


> We probably should have split this to 2 PRs, but the history should still accurately reflect credit I think.

I don't mind having the test code merged under your PR :) It makes logically sense to introduce the component with tests in the same PR

dyladan:
> > Looks good to me. I can't approve as I am the PR author 😆
> > It looks like you're only testing gauge and timer. Probably fine since all histogram-aggregated metrics should be treated the same anyway, but maybe we should also test counters.
> 
> Agree, but this is e2e testing which means not all use-cases need to be tested here. In general e2e tests should in my opinion confirm that the module in general is able to cooperate within a wider context (with processors, exporters etc.). All possible use-cases should have been covered by unit tests in upstream -> we should always aim for fast lightweight unit tests to cover most of the functionality instead of testing the component for many scenarios on a real cluster. I will try to double check the status of tests in upstream and enhance it rather directly there, so we move the maintenance of the tests to the community. WDYT?
> 
> > We probably should have split this to 2 PRs, but the history should still accurately reflect credit I think.
> 
> I don't mind having the test code merged under your PR :) It makes logically sense to introduce the component with tests in the same PR

Seems reasonable

odubajDT:
> > > Looks good to me. I can't approve as I am the PR author 😆
> > > It looks like you're only testing gauge and timer. Probably fine since all histogram-aggregated metrics should be treated the same anyway, but maybe we should also test counters.
> > 
> > 
> > Agree, but this is e2e testing which means not all use-cases need to be tested here. In general e2e tests should in my opinion confirm that the module in general is able to cooperate within a wider context (with processors, exporters etc.). All possible use-cases should have been covered by unit tests in upstream -> we should always aim for fast lightweight unit tests to cover most of the functionality instead of testing the component for many scenarios on a real cluster. I will try to double check the status of tests in upstream and enhance it rather directly there, so we move the maintenance of the tests to the community. WDYT?
> > > We probably should have split this to 2 PRs, but the history should still accurately reflect credit I think.
> > 
> > 
> > I don't mind having the test code merged under your PR :) It makes logically sense to introduce the component with tests in the same PR
> 
> Seems reasonable

Already opened a PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34041

But still adding tests

PR #233, opened 2024-07-08
[chore] refactor integration tests

bacherfl:
This pr moves code that was used for the creation of trace/metrics sinks and for the verification if expected resources have been received by them into a shared package to reduce duplications


internal/testcommon/k8s/telemetrygen.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-package k8stest // import "github.com/open-telemetry/opentelemetry-collector-contrib/internal/k8stest"
+package k8s // import "github.com/open-telemetry/opentelemetry-collector-contrib/internal/k8stest"
```
joaopgrassi, line 4: I don't have strong opinions, but I thought it was nice it was called `k8stest` as it was very clear this is "test stuff". With just `k8s` it's now not so clear. But then this is used in test code so.. maybe we gather the opinions of the others :)

bacherfl: You're right - At first i thought it might be redundant since this is part of the `testcommon` module but having it in the package name itself does make it clearer to indicate that this is only for tests - will revert it to the previous name :)

bacherfl:
converting to draft again for now as the zipkin tests that have just merged need to be adapted

joaopgrassi (approved):
thanks for improving the tests!

PR #246, opened 2024-07-23
Add initial collector dashboards 

pirgeo:


docs/dashboards/README.md
```diff
@@ -0,0 +1,163 @@
...
+      grpc: 
+        endpoint: 0.0.0.0:4317
+
+processors:
+  # transform cumulative values to deltas. 
+  cumulativetodelta/selfmon: {}
+  # (kubernetes only) retrieves kubernetes attributes for other collectors sending to this collector. See Kubernetes prerequisites below.
+  k8sattributes/selfmon: {}
+
+  # prepend 'sfm.otelcol' to all selfmon metrics - the charts in the {dashboard_name}.json file expects this prefix.
```
arminru, line 46: Might be worth pointing that out in the introduction above already for those who don't follow this snippet but already have a deployed collector and want to visualize its existing selfmon data.

docs/dashboards/README.md
```diff
@@ -0,0 +1,163 @@
...
+    metrics:
+      level: detailed
+      # export data via OTLP OTLP/grpc
+      readers:
+        - periodic:
+            interval: 20000
+            exporter:
+              otlp:
+                # the endpoint of the selfmonitoring collector. In this case, it is assumed that there is a service called `selfmon-collector` that exposes port 4317.
+                endpoint: selfmon-collector:4317
```
arminru, line 84: "selfmon-collector" is only defined in your k8s deployment below, right? Would localhost here as well and also on bare metal deployments?

pirgeo: `0.0.0.0` would only work if the collector sends to itself. For all other collectors it does not work. I guess we could do 
```suggestion
                endpoint: <location-of-your-selfmon-collector>:4317
```
or something similar to make it explicit.

pirgeo: Alternatively, we could think about distributing the `.yaml` files for Kubernetes along with the json files, but I am not sure if that wouldn't make it even more complicated

arminru: This particular section here is about the selfmon collector scraping itself specifically, right?

pirgeo: This particular section is, but I still opted for using the name of the service as the endpoint instead of `0.0.0.0` so they are consistent. This works, and if someone is copying together yaml files I believe it's easier to have them all be the same to avoid accidentally copying the wrong one.

docs/dashboards/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Architecture
+Every OpenTelemetry collector has selfmonitoring capabilities, but they need to be activated.
+Selfmonitoring data can be exported from the collector via the OTLP protocol.
+The suggested way of exporting selfmonitoring data is to run one collector dedicated to collecting and exporting the selfmonitoring data for the other running collectors, and forwarding that data to Dynatrace.
+See the diagram below for an example architecture.
+The `sidecar-collector` and `gateway-collectors` are used to send application telemetry to Dynatrace (e.g. traces, metrics, logs that the application produces).
+Both of these collectors send only their selfmonitoring data (or internal telemetry) to the `selfmon-collector`.
+The `selfmon-collector` is responsible for collecting, transforming and forwarding the internal telemetry for all collectors (including the `selfmon-collector` itself).
+Only monitoring data about the collector passes through it; no monitoring data from non-collector applicaitons is sent via the `selfmon-collector`.
```
mviitane, line 51:
```suggestion
Only monitoring data about the collector passes through it; no monitoring data from non-collector applications is sent via the `selfmon-collector`.
```

docs/dashboards/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+```mermaid
```
mviitane, line 53: This picture is difficult to read when having GitHub configured for the dark background.

docs/dashboards/README.md
```diff
@@ -0,0 +1,237 @@
...
+    metric_statements:
+      - context: metric
+        statements:
+          - set(name, Concat(["sfm.otelcol", name], "."))
+
+exporters:
+  # Inject DT_ENDPOINT and DT_API_TOKEN as environment variables. This should be the environment where the selfmonitoring data will go.
+  # See <https://docs.dynatrace.com/docs/shortlink/otel-getstarted-otlpexport> for instructions on which endpoint and token scope to use.
+  otlphttp/selfmon:
+    endpoint: "${DT_ENDPOINT}/api/v2/otlp"
```
mviitane, line 116: The endpoint and api token env variables should follow the same format as in our other use cases, for example https://docs.dynatrace.com/docs/extend-dynatrace/opentelemetry/collector/use-cases/batch

docs/dashboards/README.md
```diff
@@ -0,0 +1,237 @@
...
+    error_mode: ignore
+    metric_statements:
+      - context: metric
+        statements:
+          - set(name, Concat(["sfm.otelcol", name], "."))
+
+exporters:
+  # Inject DT_ENDPOINT and DT_API_TOKEN as environment variables. This should be the environment where the selfmonitoring data will go.
+  # See <https://docs.dynatrace.com/docs/shortlink/otel-getstarted-otlpexport> for instructions on which endpoint and token scope to use.
+  otlphttp/selfmon:
```
mviitane, line 115:
```suggestion
  otlphttp/dynatrace:
```

docs/dashboards/README.md
```diff
@@ -0,0 +1,237 @@
...
+        - periodic:
+            interval: 20000
+            exporter:
+              otlp:
+                # the endpoint of the selfmonitoring collector. In this case, it is assumed that there is a service called `selfmon-collector` that exposes port 4317.
+                endpoint: selfmon-collector:4317
+                protocol: grpc/protobuf
+                temporality_preference: delta
+
+  extensions: []
```
mviitane, line 147:
```suggestion
```

For me this line caused the following problem with helm:
`"Error: execution error at (opentelemetry-collector/templates/NOTES.txt:24:3): [ERROR] The opentelemetry-collector chart requires that the health_check extension to be included in the extension list."`

pirgeo:
Closed in favor of https://github.com/Dynatrace/dynatrace-otel-collector/pull/274

PR #249, opened 2024-07-26
fix(eecprovider): Fix authentication header format

arminru:
The EEC config provider was mistakenly sending `Api-Token: xxx` in the headers instead of `Authorization: Api-Token xxx`.

### DIFF
--- a/internal/confmap/provider/eecprovider/provider.go
+++ b/internal/confmap/provider/eecprovider/provider.go
@@ -34,7 +34,8 @@ const (
 	EECScheme SchemeType = "eec"
 )
 
-const ApiTokenHeader = "Api-Token"
+const AuthHeaderKey = "Authorization"
+const ApiTokenPrefix = "Api-Token "
 
 type provider struct {
 	caCertPath         string // Used for tests
@@ -143,7 +144,7 @@ func (p *provider) Retrieve(ctx context.Context, uri string, watcherFunc confmap
 			return nil, err
 		}
 		if cfg.authToken != "" {
-			req.Header.Add(ApiTokenHeader, cfg.authToken)
+			req.Header.Add(AuthHeaderKey, ApiTokenPrefix + cfg.authToken)
 		}
 		return req, nil
 	}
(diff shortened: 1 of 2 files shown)


odubajDT (approved):

internal/confmap/provider/eecprovider/provider.go
```diff
@@ -143,7 +144,7 @@ func (p *provider) Retrieve(ctx context.Context, uri string, watcherFunc confmap
 			return nil, err
 		}
 		if cfg.authToken != "" {
-			req.Header.Add(ApiTokenHeader, cfg.authToken)
+			req.Header.Add(AuthHeaderKey, ApiTokenPrefix + cfg.authToken)
```
pirgeo, line 147: Should we use a formatting string here maybe? Change above to: 
```golang
const ApiTokenPrefixFormat = "Api-Token %s"
```
and then here 
```suggestion
			req.Header.Add(AuthHeaderKey, fmt.Sprintf(ApiTokenPrefixFormat, cfg.authToken))
```

arminru: That would work too. @evan-bradley what's more idiomatic in Go?

evan-bradley: Neither from what I've gathered, for something simple like this it's up to the author.

evan-bradley (approved):

internal/confmap/provider/eecprovider/provider.go
evan-bradley: I have a slight preference for `fmt.Sprintf` for readability, though raw string concatenation is probably infinitesimally faster. I'm okay either way.

mowies (approved):

pirgeo (approved):

internal/confmap/provider/eecprovider/provider.go
```diff
@@ -34,7 +34,8 @@ const (
 	EECScheme SchemeType = "eec"
 )
 
-const ApiTokenHeader = "Api-Token"
+const AuthHeaderKey = "Authorization"
+const ApiTokenPrefix = "Api-Token "
```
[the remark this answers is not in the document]
arminru: Moved it to the const section.

internal/confmap/provider/eecprovider/provider.go
arminru: Updated to use the format and changed the tests to use their own literals independent from the implementation so they're decoupled.

PR #250, opened 2024-07-26
[chore]: fail testsuite when one of the tests fails

odubajDT:
## Changes

- fixes failing all-components test by adding missing statsd and zipkin receivers
- fixes statsd e2e test by adding the `//go build e2e` comment to add it to the correct CI run
- adapted Makefile to fail CI step when one of the test fails, but continue with execution of the rest of the tests, run [here](https://github.com/Dynatrace/dynatrace-otel-collector/actions/runs/10141462064/job/28038753215?pr=250)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ BIN_DIR = bin
 SRC_ROOT := $(shell git rev-parse --show-toplevel)
 
 # ALL_MODULES includes ./* dirs (excludes . dir)
-ALL_MODULES := $(shell find . -type f -name "go.mod" -not -path "./build/*" -not -path "./internal/tools/*" -exec dirname {} \; | sort | grep -E '^./' )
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -40,11 +40,13 @@ build-all: .goreleaser.yaml $(GORELEASER) $(MAIN)
...
 test: $(BIN)
+	@result=0; \
 	for MOD in $(GOMODULES); do \
 		cd $${MOD}; \
-		go test ./...; \
+		go test -v ./... || result=1; \
 		cd -; \
-	done
+	done; \
+	exit $$result;
```
mowies, line 49: this seems somewhat hacky and can probably be improved, but i guess if it works, it ain't stupid 😄

mowies (approved):

Makefile
odubajDT: it won't continue without it, didn't manage to find a better solution for now :/

pirgeo: The idea is that all tests run, even if one of them fails?

odubajDT: yes exactly

pirgeo (approved):

PR #260, opened 2024-07-31
[chore] Update changelog

evan-bradley:


mowies (commented):
i'm confused, shouldn't this have been merged before the 0.11 release?

evan-bradley:
>i'm confused, shouldn't this have been merged before the 0.11 release?

It should have, but unfortunately since I forgot to put it into the release document, I also forgot to perform this step when doing the release. I'll issue a follow-up to make sure this is taken care of.

PR #263, opened 2024-08-08
[chore] enable ARM64 linux package tests

mowies:
This PR adds linux package tests for arm64 .deb and .rpm packages using QEMU

### DIFF
--- a/.github/workflows/snapshot.yaml
+++ b/.github/workflows/snapshot.yaml
@@ -44,10 +44,9 @@ jobs:
           path: |
             dist/dynatrace-otel-collector_*_Linux_x86_64.deb
             dist/dynatrace-otel-collector_*_Linux_x86_64.rpm
+            dist/dynatrace-otel-collector_*_Linux_arm64.deb
+            dist/dynatrace-otel-collector_*_Linux_arm64.rpm
             dist/metadata.json
-#           enable ARM when github hosted ARM runners are available
-#           dist/dynatrace-otel-collector_*_Linux_arm64.deb
-#           dist/dynatrace-otel-collector_*_Linux_arm64.rpm
 
   linux-package-tests:
     name: Test Linux Packages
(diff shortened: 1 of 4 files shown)


internal/testbed/linux-services/package-tests.sh
```diff
@@ -28,6 +29,13 @@ if [[ ! -f "$PKG_PATH" ]]; then
     exit 1
 fi
 
+# translate distro x86 arch to normal name
```
pirgeo, line 32:
```suggestion
# translate distro x86 arch to podman arch name
```

mowies: but it is the generally used name i believe 😅

.github/workflows/snapshot.yaml
```diff
@@ -56,10 +55,14 @@ jobs:
...
         os: ["deb", "rpm"]
-        arch: ["x86_64"] # "arm64"] enable ARM when github hosted ARM runners are available
+        arch: ["x86_64", "arm64"]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - uses: docker/setup-qemu-action@v3
+        with:
+          platforms: arm64
+      - uses: docker/setup-buildx-action@v3
```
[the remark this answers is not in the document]
mowies: no, this should work fully transparent. Also, i specifically only setup qemu for `arm64` in line 64 here.

internal/testbed/linux-services/package-tests.sh
pirgeo: Oh, okay, I guess I am not spending enough time with folks who are discussing arch enough, I guess I am just not familiar with it 😅

mowies: at least goreleaser uses `amd64` as default and we override it 😄

.github/workflows/snapshot.yaml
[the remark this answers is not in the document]
pirgeo: Ah I guess I should look a bit more closely  🦇

PR #264, opened 2024-08-08
fix: add missing dependency to sh for RPM linux packages

mowies:
This PR synchronizes the distro linux packages with upstream updates from https://github.com/open-telemetry/opentelemetry-collector-releases/pull/617 and https://github.com/open-telemetry/opentelemetry-collector-releases/pull/620.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -67,6 +67,10 @@ nfpms:
       - apk
       - deb
       - rpm
+    overrides:
+      rpm:
+        dependencies:
+          - /bin/sh
     maintainer: "Dynatrace LLC <opensource@dynatrace.com>"
     vendor: "Dynatrace LLC"
     description: Dynatrace distribution of the OpenTelemetry Collector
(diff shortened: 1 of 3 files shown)


.goreleaser.yaml
```diff
@@ -67,6 +67,10 @@ nfpms:
       - apk
       - deb
       - rpm
+    overrides:
+      rpm:
+        dependencies:
+          - /bin/sh
```
arminru, line 73: Does this override also replace any existing dependencies not explicitly listed here, or is the default empty?

arminru (approved):

.goreleaser.yaml
mowies: The default is empty but it would override it. But since this dependency is only necessary for RPM and not for deb, that was the way to go to specifically not add it to the deb package, because it breaks the installation there

.goreleaser.yaml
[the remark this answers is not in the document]
mowies: done

pirgeo (approved):

PR #274, opened 2024-08-19
Add Dashboards for the OTel collector

pirgeo:


docs/dashboards/README.md
```diff
@@ -0,0 +1,78 @@
...
+      level: detailed
+      # set up OTLP exporter
+      readers:
+        - periodic:
+            interval: 60000
+            exporter:
+              otlp:
+                protocol: http/protobuf
+                temporality_preference: delta
+                endpoint: "${env:DT_ENDPOINT}/api/v2/otlp/v1/metrics"
```
mviitane, line 61: This should be aligned with other examples in the docs.
```suggestion
                endpoint: "${env:DT_ENDPOINT}"
```

pirgeo: In 13552d1320f1fa05636380e8f4ee6ed245fcc3f4, I changed it to use DT_URL instead of DT_ENDPOINT. That way, if we start exporting traces we can reuse the ENV VAR instead of having to add a new one just for traces.

pirgeo:
Ready to review, should only be merged/released later, though

docs/dashboards/OTel Collector self-monitoring.json
arminru, line 1: nit: Usually, this shouldn't be a problem these days, but some users might find it easier if the file name didn't contain spaces.

pirgeo: I had it like that before, but it will show that name in Dynatrace when you upload the json file. Mikko suggested using a more readable name that is nicer when displayed in the UI

arminru: Oh I see, in that case it makes sense.
I guess there's no display name property one can put in the JSON to override the file name?

pirgeo: Not that I am aware of

bacherfl (approved):

PR #270, opened 2024-08-14
Prepare release 0.12.0 (collector 0.107.0)

dyladan:
- Update version to 0.12
- Update collector and collector contrib to 0.107.0
- Add changelogs

PR #282, opened 2024-09-04
[chore] prepare release 0.13.0

mowies:
This PR prepares the changelog and manifest file for the v0.13.0 release


bacherfl (approved):

pirgeo:
In https://github.com/Dynatrace/dynatrace-otel-collector/pull/270, we also updated go.mod / go.sum files. Is is all up to date and therefore there are no gomod/gosum files updated here?

mowies:
@pirgeo in #270, the upstream collector components were updates simultaneously which caused lots of go.mod changes. In this case, that was already done in #276.
I double checked though and ran `make gotidy` which still updated some stuff, so good catch :)

pirgeo (approved):

PR #291, opened 2024-09-10
[chore] clean up testbed

odubajDT:


internal/testcommon/idutils/idutils.go
```diff
@@ -0,0 +1,22 @@
+package idutils
+
+import (
```
odubajDT, line 3: these function are needed for the upcoming traces scenarios

mowies (approved):

PR #292, opened 2024-09-11
[chore] add load tests for trace scenarios

odubajDT:


internal/testbed/load/tests/trace_test.go
```diff
@@ -0,0 +1,197 @@
...
+				},
+				resourceSpec: testbed.ResourceSpec{
+					ExpectedMaxCPU: 30,
+					ExpectedMaxRAM: 120,
+				},
+				attrCount:       10,
+				attrSizeByte:    50,
+				attrKeySizeByte: 50,
+			},
+			traceLimitProcessors,
```
mowies, line 115: i don't think we need actually limited tests

mowies (approved):

PR #297, opened 2024-09-16
[chore] add load tests for logs scenarios 

bacherfl:
Note: Right now the syslog data sender sends invalid messages that cause an error within the syslog receiver. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35201 fixes this


internal/testbed/go.mod
```diff
@@ -5,12 +5,12 @@ go 1.22.6
 require (
 	github.com/Dynatrace/dynatrace-otel-collector/internal/testcommon v0.0.0
 	github.com/google/uuid v1.6.0
-	github.com/open-telemetry/opentelemetry-collector-contrib/testbed v0.108.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/testbed v0.109.1-0.20240917213807-5cb166da005b
```
bacherfl, line 8: Using the latest commit on main for the testbed for now, as this includes https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35201 which is needed for the syslog test.

PR #299, opened 2024-09-16
[chore] prepare release 0.14.0

odubajDT:


dyladan (approved):

internal/tools/go.mod
```diff
@@ -7,7 +7,6 @@ toolchain go1.23.1
 require (
 	github.com/goreleaser/goreleaser/v2 v2.3.1
 	github.com/jstemmer/go-junit-report v1.0.0
-	github.com/jstemmer/go-junit-report/v2 v2.1.0
```
dyladan, line 10: Why was this removed?

mowies (approved):

internal/tools/go.mod
odubajDT: `go mod tidy` was executed just to clean up unused dependencies, this dependency is not used right now at all, but we will try to introduce it via https://github.com/Dynatrace/dynatrace-otel-collector/pull/298

It's has zero importance for the release

pirgeo (approved):

PR #320, opened 2024-10-09
[chore] use bigger runner for load tests

mowies:
This PR changes the runner for our load tests to our new github hosted larger runner with 16 core.

### DIFF
--- a/.github/workflows/load-tests.yml
+++ b/.github/workflows/load-tests.yml
@@ -57,7 +57,9 @@ jobs:
           path: ./bin/*
 
   loadtest:
-    runs-on: ubuntu-24.04
+    runs-on:
+      group: "Dynatrace-OTel-Runners"
+      labels: "Linux-x64-16-64GB-Runner"
     needs: [setup-environment]
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/load-tests.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           path: ./bin/*
 
   loadtest:
-    runs-on: ubuntu-24.04
+    runs-on: "Linux-x64-16-64GB-Runner"
```
arminru, line 60: If we know the image version (ubuntu-latest or ubuntu-24.04?) we could document it in a comment.

mowies: done

odubajDT:
what about the combined scenario? Do you think we need a bigger runner for that as well?

mowies:
good point! yes we should also use the runner there

mowies:
the combined load test is now also moved to the larger runner and it only runs on pushes to main branch same as for the rest of the load tests

bacherfl (approved):

PR #331, opened 2024-10-18
[chore] introduce testing workflow for filtering scenarios

odubajDT:
## Changes

- Introduces a base scenario for filtering use-cases
- refactors helper functions


Makefile
```diff
@@ -49,7 +49,7 @@ test: $(BIN)
 	@result=0; \
 	for MOD in $(GOMODULES); do \
 		cd $${MOD}; \
-		go test -v ./... || result=1; \
+		go test -v -timeout 90s ./... || result=1; \
```
odubajDT, line 52: the reason for this change is that the filtering scenario takes quite some amount of time when executed as table tests. This will be adapted accordingly, once we start to implement real testing scenarios for filtering

internal/testcommon/testutil/testutil.go
```diff
@@ -182,3 +183,118 @@ func ReplaceDynatraceExporterEndpoint(cfg string, exporterPort int) string {
...
+	processors map[string]string,
+	extensions map[string]string,
+) string {
+
+	pprofString := fmt.Sprintf(`
+  pprof:
+    save_to_file: %v/cpu.prof
+`, resultDir)
+
+	extensions["pprof"] = pprofString
```
bacherfl, line 201:
```suggestion
        if extensions == nil {
          extensions = map[string]string{}
        }
	extensions["pprof"] = pprofString
```

internal/testbed/integration/filtering_scenario.go
```diff
@@ -0,0 +1,76 @@
...
+
+var (
+	defaultProcessors = map[string]string{
+		"memory_limiter": `
+  memory_limiter:
+    check_interval: 1s
+    limit_percentage: 100
+`,
+		"batch": `
+  batch:
```
bacherfl, line 20: do we need the batch processor here? As we are sending only a small number of data items I think we might slow down the tests as the queue will not be full until the 10 second timeout has been reached

odubajDT: the scenarios added to the test are just to showcase how the tests should be build, we can definitely remove the scenarios that are currently present, since they are not filtering anything. I just wanted to proof that the setup works and can be therefore used for filtering testing

bacherfl (approved):

PR #332, opened 2024-10-21
[chore] adapt prometheus load test limits

bacherfl:


internal/testbed/load/tests/metrics_test.go
```diff
@@ -254,7 +254,7 @@ func TestPrometheusMetric(t *testing.T) {
 			extendedLoadOptions: ExtendedLoadOptions{
 				resourceSpec: testbed.ResourceSpec{
 					ExpectedMaxCPU: 200,
-					ExpectedMaxRAM: 370,
+					ExpectedMaxRAM: 550,
```
mowies, line 257: is there maybe an option to decrease the memory checking granularity to allow for spikes?

bacherfl: there is a `MaxConsecutiveFailures` option - I will try to set this to `2` and reduce the limits by a bit again and see how this behaves

mowies (approved):

PR #337, opened 2024-10-22
Introduce redaction example of credit card numbers

odubajDT:
## Changes

- introduced configuration for redaction of credit card number using the redaction processor
- adapted the validators to assert full data types
- adapted config example tests according to validators

## Follow up

- introduce configuration for transform processor to mask credit cards numbers (all numbers except the last 4)


internal/testbed/integration/metrics_validation.go
```diff
@@ -45,18 +43,23 @@ func assertExpectedMetricsAreInReceived(t *testing.T, expected, actual []pmetric
 				metrics := scopeMetrics.At(j).Metrics()
 				for k := 0; k < metrics.Len(); k++ {
 					actualMetric := metrics.At(k)
-					assert.Contains(t,
+					hasEntry := assert.Contains(t,
```
bacherfl, line 46: nit: we could use `require.Contains` here so we don't have to do the check for `hasEntry` below

bacherfl (approved):

mowies (approved):

config_examples/filtering_creditcards.yaml
```diff
@@ -0,0 +1,37 @@
...
+      grpc:
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
+
+processors:
+  redaction:
+    allow_all_keys: true
+    allowed_keys:
+      - credit_card
```
arminru, line 13: Would this also work without these two lines since we already have `allow_all_keys: true` which should implicitly allow `credit_card` and only blocked `blocked_values`?

From https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/redactionprocessor/README.md:

    # allow_all_keys is a flag which when set to true, which can disables the
    # allowed_keys list. The list of blocked_values is applied regardless. If
    # you just want to block values, set this to true.
    allow_all_keys: false
    # allowed_keys is a list of span/log/datapoint attribute keys that are kept on the span/log/datapoint and
    # processed. The list is designed to fail closed. If allowed_keys is empty,
    # no attributes are allowed and all span attributes are removed. To
    # allow all keys, set allow_all_keys to true.
    allowed_keys:

odubajDT: oh good catch!, my bad, there should be `allow_all_keys: false`. Will adapt it in a follow up PR

arminru: --> #340

PR #342, opened 2024-10-24
Add user data filtering example config

bacherfl:


odubajDT (approved):

pirgeo (approved):

config_examples/filtering_user_data.yaml
```diff
@@ -0,0 +1,55 @@
...
+          - set(attributes["user.name"], "****")
+          - set(attributes["user.full_name"], "****")
+          - set(attributes["user.email"], "****")
+    log_statements:
+      - context: log
+        statements:
+          - set(attributes["user.id"], "****")
+          - set(attributes["user.name"], "****")
+          - set(attributes["user.full_name"], "****")
+          - set(attributes["user.email"], "****")
```
mowies, line 37: you could use YAML anchors here to have less duplication

bacherfl: good suggestion, thanks - adapted :)

PR #340, opened 2024-10-23
Introduce masking example of credit card numbers

odubajDT:
## Changes

- add more credit card examples to the redaction test
- add filtering test using transform processor
- add example configuration for filtering using transform processor
- small fix of filtering example with redaction processor

PR #343, opened 2024-10-24
Introduce masking and redaction example of international IBANs

odubajDT:
## Changes

- introduce example config to mask IBANs with transform processor
- introduce example config to redact IBANs with redaction processor
- introduce integration tests for both configs

### DIFF
--- a/config_examples/masking_iban.yaml
+++ b/config_examples/masking_iban.yaml
@@ -8,19 +8,22 @@ receivers:
 
 processors:
   transform:
-  error_mode: ignore
-  trace_statements:
-    - context: span
-      statements:
-        - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
-  metric_statements:
-    - context: datapoint
-      statements:
-        - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
-  log_statements:
-    - context: log
-      statements:
-        - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
+    error_mode: ignore
+    trace_statements:
+      - context: span
+        statements:
+          - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
+      - context: spanevent
+        statements:
+          - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
+    metric_statements:
+      - context: datapoint
+        statements:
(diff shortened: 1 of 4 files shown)


config_examples/masking_iban.yaml
```diff
@@ -0,0 +1,50 @@
...
+      statements:
+        - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
+  metric_statements:
+    - context: datapoint
+      statements:
+        - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
+  log_statements:
+    - context: log
+      statements:
+        - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
```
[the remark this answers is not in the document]
odubajDT: I was thinking about this as well, but YAML does not support any definition of constants. I tried to do some magic with [anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/). But it does not seem to me as a good idea for example configuration.

internal/testbed/integration/filtering_test.go
```diff
@@ -96,6 +96,209 @@ func TestFilteringCreditCard(t *testing.T) {
...
+    metric_statements:
+      - context: datapoint
+        statements:
+          - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
+    log_statements:
+      - context: log
+        statements:
+          - replace_all_patterns(attributes, "value", "^([A-Z]{2})[0-9]{2}((\\s*[A-Z0-9]){4,26})\\s*((\\s*[A-Z0-9]){4})$", "$$1 **** $$4")
+`,
+	}
```
arminru, line 227: Can this be taken directly from `config_examples\masking_iban.yaml` rather than duplicating and inlining it?

This way, it would always be kept in sync and changes to the YAML file would be tested directly while we wouldn't need to think of reflecting changes here as well.

Also, the invalid config pointed out in https://github.com/Dynatrace/dynatrace-otel-collector/pull/343#discussion_r1814893699 would've been spotted immediately as the test would've failed.

odubajDT: we should be able to extract only the processors part from the config to avoid rewriting the whole filtering scenario, since it's already used in multiple tests (most of them non-merged, but ready). I will try to create a wrapper, that will do the extraction from the example config to a map, since in our cases, receiver/exporters are not important here. Do you agree?

odubajDT: adapted the functionality [here](https://github.com/Dynatrace/dynatrace-otel-collector/pull/340/commits/000067fbdb38c89f339187c83cba6a6abb12d6a4) as part of a different PR (that should be merged before this one), please have a look there if you have any objections, otherwise I will proceed with merging

arminru: That change looks good to me. Thanks!

joaopgrassi (approved):

pirgeo (commented):
Do we have some sort of documentation about the difference of redacting and masking? It's clear to me now, but it might be helpful to have a readme or similar that says "use either one or the other, you don't need both" or something along those lines

pirgeo (approved):

odubajDT:
> Do we have some sort of documentation about the difference of redacting and masking? It's clear to me now, but it might be helpful to have a readme or similar that says "use either one or the other, you don't need both" or something along those lines

Not now, but I can sync with Alex regarding this and include it to the example config page, since we will probably include both examples there.

PR #353, opened 2024-10-30
[chore] refactor filtering integration test scenario

odubajDT:


internal/testcommon/testutil/testutil.go
```diff
@@ -302,3 +303,30 @@ service:
 		receiver.ProtocolName(),
 	)
 }
+
+type ProcessorConfig struct {
+	Processors map[string]any `yaml:"processors"`
+}
+
+func extractProcessorsFromYAML(yamlStr []byte) (map[string]string, error) {
```
odubajDT, line 311: not used in current code, but good helper function for potential future tests, where we want to test only processors parts

bacherfl (approved):

PR #357, opened 2024-11-05
Extend config examples with log body filtering

mowies:
This PR adds credit card and IBAN filtering for log bodies to the config examples and tests


bacherfl (approved):

odubajDT (approved):

config_examples/masking_logbody.yaml
```diff
@@ -0,0 +1,36 @@
...
+      Authorization: "Api-Token ${env:API_TOKEN}"
+
+processors:
+  transform/cc:
+    error_mode: ignore
+    trace_statements:
+    log_statements:
+      - context: log
+        statements:
+          - replace_pattern(body, "3\\s*[47](\\s*[0-9]){9}((\\s*[0-9]){4})$", "**** $$2")
```
pirgeo, line 19: Doesnt the `$` at the end mean that this only identifies numbers if the log message ends with the credit card number? I.e.: 
```
2024/11/08 09:54 [paymentservice] This is a log message that contains this credit card number: 2367 8901 2345 6789
```
Would be redacted but for this it wouldn't work:
```
2024/11/08 09:54 [paymentservice] Customer ABC used AMEX 2367 8901 2345 6789 to pay for transaction 12345.
```

internal/testbed/integration/filtering_test.go
```diff
@@ -155,6 +155,163 @@ func TestFilteringCreditCard(t *testing.T) {
 	}
 }
 
+func TestFilteringLogBody(t *testing.T) {
+	logsCCPlain := []string{
+		"card_master_spaces1 2367 8901 2345 6789",
```
pirgeo, line 160: Maybe we can add test cases for `<text> <number> <text>` or just `<number> <text>` (see my other comment)

config_examples/masking_logbody.yaml
mowies: yes, that's true, I guess I can adapt it a bit to make the regex more general

internal/testbed/integration/filtering_test.go
mowies: done

config_examples/masking_logbody.yaml
mowies: done

pirgeo (approved):

PR #367, opened 2024-11-08
[chore] prepare release 0.18.0

bacherfl:


CHANGELOG.md
```diff
@@ -4,6 +4,74 @@
...
+---
+
+</details>
+
+#### Dynatrace distribution changelog:
+
+### 💡 Enhancements 💡
+
+- `transformprocessor`: Extend the user data filtering example with a statement for masking the `client.address` attribute. (#356)
+- `transformprocessor`: Extend the credit card filtering example with log body filtering (#357)
```
mowies, line 72: [nitpick] I forgot to add that also IBAN log body filtering is included here

mowies (approved):

CHANGELOG.md
bacherfl: thanks, I have adapted the not to also reflect that

mowies (approved):

pirgeo:
Should we run `make gotidy` before releasing?

bacherfl:
> Should we run `make gotidy` before releasing?

good catch, I will also add a note about this to the release check list so we don't forget it in the future

pirgeo (approved):

odubajDT (approved):

PR #373, opened 2024-11-18
[chore] cleanup and refactor testing workflows

odubajDT:


internal/testbed/chainsaw/syslog-receiver/00-assert.yaml
odubajDT, line 1: test not needed, as it was not used and substituted by https://github.com/Dynatrace/dynatrace-otel-collector/blob/main/internal/testbed/integration/config_examples_test.go#L575

internal/testbed/chainsaw/prometheus-node-exporter/00-assert.yaml
odubajDT, line 1: test not needed as it is was not used and subtituted by https://github.com/Dynatrace/dynatrace-otel-collector/tree/main/internal/testbed/integration/prometheus

bacherfl (approved):

.github/actions/create-cluster/action.yaml
```diff
@@ -10,22 +10,55 @@ inputs:
     required: false
     description: "Kubernetes version that should be used"
     # renovate: datasource=github-releases depName=kubernetes/kubernetes
-    default: "v1.27.3"
+    default: "v1.27.11"
```
mowies, line 13: this is not actually picked up by renovate...

.github/workflows/e2e.yaml
```diff
@@ -160,48 +120,10 @@ jobs:
...
-
-      - name: Load Docker image
-        run: |
-          docker load --input /tmp/dynatrace-otel-collector.tar
-
-      - name: Load Image into Kind
-        shell: bash
-        run: |
-          kind load docker-image dynatrace-otel-collector:e2e-test --name kind
+          runtime_tag:
```
mowies, line 126: this still needs a value since it's a required input

.github/actions/create-cluster/action.yaml
mowies: not necessarily something to be solved here though

odubajDT: will be handled in follow-up

mowies (approved):

PR #400, opened 2024-12-06
[chore] increase limits for load tests

bacherfl:
According to the recent results https://dynatrace.github.io/dynatrace-otel-collector/benchmarks/loadtests/ there was no significant or sudden increase in the memory consumption, but it was always rather close to the specified limits

### DIFF
--- a/internal/testbed/load/tests/metrics_test.go
+++ b/internal/testbed/load/tests/metrics_test.go
@@ -55,7 +55,7 @@ func TestMetric10kDPS(t *testing.T) {
 				},
 				resourceSpec: testbed.ResourceSpec{
 					ExpectedMaxCPU: 60,
-					ExpectedMaxRAM: 125,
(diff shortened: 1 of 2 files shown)


odubajDT (approved):

pirgeo:
Do you think it makes sense to add the maximum specified for the test into the graphs as well? So we can see how close we are to hitting the cutoff point?

bacherfl:
> Do you think it makes sense to add the maximum specified for the test into the graphs as well? So we can see how close we are to hitting the cutoff point?

that's a good idea - I'd look into that in a separate PR, as I need to figure out how to best get this data into the result json file generated by the testing library

pirgeo:
Sounds good, definitely should go in a separate PR. Since you mentioned "we've been pretty close to the limits" I looked at the overview page and realized I have no idea what the limits actually are 😄

pirgeo (approved):

PR #439, opened 2025-01-28
[chore] add cli tool to ingest otlp data to the collector

bacherfl:


internal/data-ingest-cli/main.go
```diff
@@ -0,0 +1,57 @@
...
+	"flag"
+	"fmt"
+	"github.com/Dynatrace/dynatrace-otel-collector/internal/data-ingest-cli/commands/otlpjson"
+	"log"
+)
+
+func main() {
+	// Define the CLI arguments
+	inputFile := flag.String("input-file", "", "Path to the input file containing input data")
+	collectorURL := flag.String("collector-url", "", "URL of the OpenTelemetry collector")
```
pirgeo, line 14: Should the default be `http://localhost:4317` for the collector URL?

internal/data-ingest-cli/README.md
```diff
@@ -0,0 +1,46 @@
...
+## Features
+
+- Read OTLP JSON data from a file: The tool can read OTLP data from a specified file containing data in different formats and send it to an OpenTelemetry collector.
+ Supported formats are:
+  - OTLP Json
+  - Systemd
+  - Syslog
+
+- Receive OTLP data via an endpoint: The tool can receive OTLP data through an endpoint and store the received payload in a specified file in OTLP Json format.
+
```
pirgeo, line 21:
```suggestion

- Send the data read from the JSON file to an OTLP endpoint via gRPC.
```

internal/data-ingest-cli/README.md
```diff
@@ -0,0 +1,46 @@
...
+- `--collector-url`: The URL of the OpenTelemetry collector.
+- `--output-file`: The file in which to store the received data.
+- `--receiver-port`: The port of the OTLP receiver created to act as a sink for the collector.
+
+## Example Commands
+
+1. Send OTLP JSON data to a collector:
+
+```
+./data-ingest --input-file=data.json --input-format=otlpjson --collector-url=http://collector.example.com:4317 --output-file=received.json --receiver-port=4319
```
pirgeo, line 45: This does all at once, right? It receives, stores in a file, but also sends to the collector at `http://collector.example.com:4317/`

internal/data-ingest-cli/main.go
```diff
@@ -0,0 +1,57 @@
...
+	outputFile := flag.String("output-file", "", "Path to the file where received OTLP data will be stored")
+	inputFormat := flag.String("input-format", "otlp-json", "Input format (options: 'otlp-json', 'syslog', 'systemd')")
+	otlpSignalType := flag.String("otlp-signal-type", "", "OTLP signal type (options: 'logs', 'traces', 'metrics')")
+	receiverPort := flag.Int("receiver-port", 0, "OTLP Receiver port. If set, the tool will open a grpc server on the specified port to receive data and store it in an output file")
+
+	// Parse the CLI arguments
+	flag.Parse()
+
+	// Validate required arguments
+	if *collectorURL == "" {
```
pirgeo, line 24: I know this is contrary to having a default for the collector URL, but should this fail if there is no collector URL? I can see this tool being deployed somewhere simply to extract some test data for reproducible problem cases, and you might not want to send data out in that case

internal/data-ingest-cli/README.md
bacherfl: yes, so basically the flow is as follows:

- Read data from the input file
- Send that data to the collector
- Act as a sink for the collector to send the processed data via the otlp exporter
- Store the received data in a file

This should allow us to easily compare the input and output files and how the ingested data looks after running through the collector's pipeline

internal/data-ingest-cli/main.go
bacherfl: yes good point - i will check and see if i can refine the parameter validation a bit more and set some defaults where applicable

bacherfl: actually, right now the tool is not really designed to run as only a OTLP receiver, as it will terminate as soon as it has received the precessed data item it has sent previously.
However, it might be useful to also support the use case you mentioned, since the dt-collector does not include the file exporter. We could think of adding this in a follow up PR, wdyt?

internal/data-ingest-cli/README.md
```diff
@@ -0,0 +1,47 @@
...
+making it easier to diagnose and resolve problems related to OTLP data handling.
+
+*Note:* This is a developer tool used mainly for debugging, and not intended for being used within a production environment.
+
+## Features
+
+- Read OTLP JSON data from a file: The tool can read OTLP data from a specified file containing data in different formats and send it to an OpenTelemetry collector.
+ Supported formats are:
+  - OTLP Json
+  - Systemd
```
odubajDT, line 17: You probably mean statsd no ?
```suggestion
  - Statsd
```

internal/data-ingest-cli/README.md
```diff
@@ -0,0 +1,47 @@
...
+```
+go build -o data-ingest
+```
+
+## Usage
+
+The tool accepts the following input parameters:
+
+- `--input-file`: The name of the input file to read data from.
+- `--input-format`: The input format of the ingested data (options: `otlpjson`, `syslog`, `systemd`).
```
odubajDT, line 36:
```suggestion
- `--input-format`: The input format of the ingested data (options: `otlpjson`, `syslog`, `statsd`). 
```

internal/data-ingest-cli/main.go
```diff
@@ -0,0 +1,57 @@
...
+		if err != nil {
+			log.Fatalf("could not execute command: %s", err.Error())
+		}
+		if err := cmd.Do(context.Background()); err != nil {
+			log.Fatalf("could not execute command: %s", err.Error())
+		}
+	case "syslog":
+		// Handle reading from syslog and sending to collector
+		fmt.Println("Reading from syslog and sending to collector...")
+	case "systemd":
```
odubajDT, line 52:
```suggestion
	case "statsd":
```

internal/data-ingest-cli/README.md
bacherfl: good catch, thanks, that is indeed what i meant

internal/data-ingest-cli/main.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/Dynatrace/dynatrace-otel-collector/internal/data-ingest-cli/commands/otlpjson"
+	"log"
+)
+
+func main() {
+	// Define the CLI arguments
+	inputFile := flag.String("input-file", "", "Path to the input file containing input data")
+	collectorURL := flag.String("collector-url", "localhost:4317", "URL of the OpenTelemetry collector")
+	outputFile := flag.String("output-file", "", "Path to the file where received OTLP data will be stored")
+	inputFormat := flag.String("input-format", "otlp-json", "Input format (options: 'otlp-json', 'syslog', 'systemd')")
```
odubajDT, line 16:
```suggestion
	inputFormat := flag.String("input-format", "otlp-json", "Input format (options: 'otlp-json', 'syslog', 'statsd')")
```

internal/data-ingest-cli/main.go
```diff
@@ -0,0 +1,57 @@
...
+			log.Fatalf("could not execute command: %s", err.Error())
+		}
+		if err := cmd.Do(context.Background()); err != nil {
+			log.Fatalf("could not execute command: %s", err.Error())
+		}
+	case "syslog":
+		// Handle reading from syslog and sending to collector
+		fmt.Println("Reading from syslog and sending to collector...")
+	case "statsd":
+		log.Println("Reading from systemd and sending to collector...")
```
odubajDT, line 53:
```suggestion
		log.Println("Reading from statsd and sending to collector...")
```

odubajDT (approved):

PR #460, opened 2025-02-13
[chore] add host metrics config example

mowies:
Signed-off-by: Moritz Wiesinger <moritz.wiesinger@dynatrace.com>


internal/data-ingest-cli/main.go
```diff
@@ -47,6 +52,8 @@ func main() {
 	case "otlp-json":
 		fmt.Println("Reading otlpjson data and sending it to collector...")
 		cmd, err := otlpjson.New(otlpjson.Config{
+			SendData:     *sendData,
```
bacherfl, line 55: instead of checking in the command implementations, should we rather check for `sendData` here and add a dedicated command for just receiving data?

odubajDT (approved):

bacherfl (approved):

PR #475, opened 2025-02-24
Add resource detection config example and integration test

mowies:


bacherfl (approved):
LGTM, i think the failing EEC e2e test is unrelated to these changes, but please retry that one

internal/testbed/integration/resource-detection/testdata/collector/enrichment-file.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Name }}-enrichment-file
+  namespace: e2eresourcedetection
+data:
+  dt_host_metadata.properties: |
+    dt.entity.host=my-host-from-properties
+    host.name=my-host-from-properties
```
arminru, line 9: For the matter of testing it might make more sense to use two different values here so we can tell that it takes the right one each and doesn't mix them up or only takes the same for both.

PR #491, opened 2025-03-11
[chore] Use lychee to check links

mowies:
This PR replaces markdown-link-check with lychee, since md-link-check had long standing unresolved issues in the past and trust into the project is not there anymore.

Upstream reference: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37100


bacherfl (approved):

.github/workflows/changelog.yml
```diff
@@ -87,12 +85,8 @@ jobs:
...
-        run: npm install -g markdown-link-check@${{ env.MARKDOWN_LINK_CHECK_VERSION }}
-
-      - name: Run markdown-link-check
-        run: |
-          markdown-link-check \
-            --verbose \
-            --config .markdown_link_check_config.json \
-            changelog_preview.md
+      - name: Run Lychee
+        uses: lycheeverse/lychee-action@v2.3.0
```
odubajDT, line 89: what about renovate updates?

mowies: since this is just a github action, renovate will automatically pick it up. let's see how it goes and if we need to, we can pin down the CLI version and then put that under renovate with usual annotation comment

PR #507, opened 2025-03-18
[chore] add Lychee link scanning for the complete repo

mowies:
This PR adds lychee link scanning on PRs and on push to main for the complete repo.
It also runs the link checker every monday morning and will file an issue if broken links come up.

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -1,0 +1,44 @@
+# This action requires that any PR targeting the main branch should touch at
+# least one CHANGELOG file. If a CHANGELOG entry is not required, add the "Skip
+# Changelog" label to disable this action.
+
+name: 'Check Links'
+
+on:
+  pull_request:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+  schedule:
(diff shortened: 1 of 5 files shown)


.github/workflows/check-links.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          fail: ${{ github.event_name != 'schedule' }}
+
+      - name: Create Issue From File
+        if: steps.lychee.outputs.exit_code != 0 && github.event_name == 'schedule'
+        uses: peter-evans/create-issue-from-file@e8ef132d6df98ed982188e460ebb3b5d4ef3a9cd # v5.0.1
+        with:
+          title: Lychee found some broken links
+          content-filepath: ./lychee/out.md
+          labels: bug
```
arminru, line 44: I'd also be fine with the build failing on the next PR (or main if we keep the schedule) in case you'd like to avoid pulling in another GH workflow action.

arminru (approved):

.github/workflows/check-links.yml
mowies: no i think that's fine, better to get notified when something is wrong

PR #527, opened 2025-03-24
[chore] add E2E test for self monitoring metrics

odubajDT:


config_examples/self-monitoring.yaml
```diff
@@ -0,0 +1,43 @@
+extensions:
```
arminru, line 1:
```suggestion
# follows https://docs.dynatrace.com/docs/shortlink/otel-collector-self-monitoring#read-attributes-from-the-container-environment
extensions:
```
So we know where to update this file from later.

config_examples/self-monitoring-k8s-enrich.yaml
```diff
@@ -0,0 +1,99 @@
+extensions:
```
arminru, line 1:
```suggestion
# follows https://docs.dynatrace.com/docs/shortlink/otel-collector-self-monitoring#enrich-data-using-the-k8sattributes-processor
extensions:
```
So we know where to update this file from later.

internal/testbed/integration/self-monitoring/e2e_test.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}()
+
+	wantEntries := 1 // Minimal number of metrics to wait for.
+	oteltest.WaitForMetrics(t, wantEntries, metricsConsumer)
+
+	metrics := pmetric.NewMetrics()
+	resource := metrics.ResourceMetrics().AppendEmpty()
+
+	scope1 := resource.ScopeMetrics().AppendEmpty()
```
bacherfl, line 120: can we use the same approach for generating the expected metrics as in e.g. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/55b80b5efaa81bbce099f0d3f7e5e948fa870799/receiver/k8sclusterreceiver/e2e_test.go#L167?

This would allow us to detect any naming changes in any of the self monitoring metrics and warn us early enough about breaking changes. This would also make it easier to verify if other properties like the temporality preference have been set correctly

odubajDT: I'll have a look, let's see if we can use this

odubajDT: ok, spent some time on this and for the current state this is not the best idea: one problem with the generation of the results are comparison of exemplars, which are part of Sum metrics, but that's doable by contributing upstream to add an ignorer for exemplars in the pmetrictest options library.

Another problem is (only for the k8s-enrich test, where we actually push some data to the pipeline) that collector is generating big amount of metrics, where basically in each iteration (the longer you wait) you get more and more metrics, therefore is almost impossible to actually compare, since there are always newer and newer metrics appearing, basically the longer you wait, which makes completely sense for observability purposes, but makes it unable to compare with expected output. The best thing we can make here is to check if the metrics, that are important for us are present.

The generation library is useful, but in my opinion in scenarios, where you have the power to customize the input for the collector pipeline. For an automatically generated metrics, it's not the best solution

bacherfl: ok thank you for looking into this. In this case we can use the current tests as a start and look into how we can best add more detailed checks as a follow up

bacherfl (approved):

PR #571, opened 2025-05-15
[chore] Prepare v0.3.0 release

evan-bradley:


dyladan (approved):

pirgeo:
Would it be 0.30.0 (not 0.3.0)?

evan-bradley:
>Would it be 0.30.0 (not 0.3.0)?

Good catch! I'll issue a follow-up.

PR #581, opened 2025-05-20
[chore] add integration test k8s receivers combined scenario

odubajDT:


config_examples/k8scombined-gateway.yaml
```diff
@@ -0,0 +1,125 @@
+extensions:
+  health_check:
+    endpoint: 0.0.0.0:13133
+
+receivers:
```
mowies, line 5: i think it would be better from a data flow persepective to have the gateway collector receive and forward the traffic from the agents. So i would add the otlp receiver here and add a second forwarding pipeline for the data

odubajDT: adapted

internal/testbed/integration/k8scombined/e2e_test.go
```diff
@@ -0,0 +1,429 @@
...
+        - kubeletstats
+      processors:
+        - filter
+        - k8sattributes
+      exporters:
+        - otlp`,
+			fmt.Sprintf(`
+  otlphttp:
+    endpoint: http://%s:4321
+    headers:
```
bacherfl, line 211: i guess the headers section is not needed here?

internal/testbed/integration/k8scombined/e2e_test.go
```diff
@@ -0,0 +1,429 @@
...
+	require.NoError(t, err)
+
+	require.EventuallyWithT(t, func(tt *assert.CollectT) {
+		assert.NoError(tt, pmetrictest.CompareMetrics(expected, metricsConsumerGateway.AllMetrics()[len(metricsConsumerGateway.AllMetrics())-1],
+			defaultOptions...,
+		),
+		)
+	}, 3*time.Minute, 1*time.Second)
+
+	t.Logf("Gateway metrics checked successfully")
```
bacherfl, line 412: if i understand the test correctly, both collectors (agent and gateway) are tested separately, as the otlp exporter of the agent is modified to send to the local sink instead of `otelcolsvc:4317`. This definitely makes sense, but I was wondering if we should also add a test where keep the agent collectors' exporter configuration unchanged to send data to the gateway, and then only check what we get from the gateway collector?
If yes, I think that should be a follow up though, so I would not include that in this PR

bacherfl (approved):
lgtm with one nit and a suggestion for a potential follow up

internal/testbed/integration/k8scombined/e2e_test.go
```diff
@@ -0,0 +1,429 @@
...
+      processors:
+        - filter
+        - k8sattributes
+      exporters:
+        - otlp`,
+			fmt.Sprintf(`
+  otlphttp:
+    endpoint: http://%s:4321
+    headers:
+      Authorization: "Api-Token "
```
odubajDT, line 212:
```suggestion
```

internal/testbed/integration/k8scombined/e2e_test.go
```diff
@@ -0,0 +1,427 @@
...
+        - k8sobjects
+      processors:
+        - transform
+      exporters: 
+        - otlphttp`,
+			fmt.Sprintf(`
+  otlphttp/metrics:
+    endpoint: http://%s:4320
+    headers:
+      Authorization: "Api-Token "
```
odubajDT, line 316:
```suggestion
```

internal/testbed/integration/k8scombined/e2e_test.go
```diff
@@ -0,0 +1,427 @@
...
+  otlphttp/logs:
+    endpoint: http://%s:4319
+    headers:
+      Authorization: "Api-Token "
```
odubajDT, line 320:
```suggestion
```

internal/testbed/integration/k8scombined/e2e_test.go
bacherfl: that makes sense, thanks for elaborating!

config_examples/k8scombined-gateway.yaml
```diff
@@ -0,0 +1,137 @@
...
+        enabled: true
+
+processors:
+  cumulativetodelta:
+  transform:
+    error_mode: ignore
+    metric_statements:
+      - context: resource
+        statements:
+          - set(attributes["k8s.cluster.name"], "k8s-testing-cluster")
```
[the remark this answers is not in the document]
odubajDT: Fair point, we can add an env variable and point out in the docs that this variable needs to be set to the cluster name (or whatever the user wants to). Would this solution work?

There is actually no standard possibility to determine the cluster name from inside the cluster, so this needs to be set manually (not without escalating priviledges). Also, every provider (AWS, Google cloud etc.) can have a different possibilities for this

config_examples/k8scombined-gateway.yaml
```diff
@@ -0,0 +1,137 @@
...
+service:
+  extensions:
+    - health_check
+  pipelines:
+    metrics/forward:
+      receivers:
+        - otlp
+      processors:
+        - transform
+        - cumulativetodelta
```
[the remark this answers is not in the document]
pirgeo: I was thinking if it was possible to have the cluster scope metrics exported as Delta by default (at the producer already), then we would not need it. If that's not possible, then leaving it as is is fine!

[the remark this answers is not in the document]
odubajDT: There isn't any option like that in the [k8scluster receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/k8sclusterreceiver), I guess it's not possible

PR #597, opened 2025-06-17
[chore] Move to GoReleaser Pro

mowies:
This PR:
- adapts the makefile and tools so that goreleaser-pro is installed from github releases and not through go modules since that's not possible for the pro version. This also covers when the wrong goreleaser version is installed
- exchanges the goreleaser usages in snapshot and releases pipelines to use the goreleaser-action instead of makefile targets
- removes the installation step for cyclone-dx since SBOM generation is skipped for snapshot builds

### DIFF
--- a/.github/workflows/snapshot.yaml
+++ b/.github/workflows/snapshot.yaml
@@ -44,7 +44,7 @@ jobs:
       - uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
         if: matrix.goarch != 'amd64'
         with:
-          platforms: arm64,ppc64le,s390x
+          platforms: ${{ matrix.goarch }}
 
       - uses: docker/setup-buildx-action@18ce135bb5112fa8ce4ed6c17ab05699d7f3a5e0 # v3.11.0
 


Makefile
```diff
@@ -65,18 +68,50 @@ clean-tools:
...
+URL := https://github.com/goreleaser/goreleaser-pro/releases/download/$(GORELEASER_PRO_VERSION)/$(BINARY_NAME)
+
+install-goreleaser-pro:
+	echo 'Installing GoReleaser Pro...'; \
+	if [ "v$(GORELEASER_ACTUAL_VERSION)" = "$(GORELEASER_PRO_VERSION)" ]; then \
+	  	echo "GoReleaser is already installed with the correct version, moving on..."; \
+	else \
+		echo "Downloading $(BINARY_NAME) from $(URL)..."; \
+		curl -L $(URL) -o $(BINARY_NAME); \
+		if [ "$(EXT)" = "zip" ]; then unzip -o "$(BINARY_NAME)"; else tar -xzf "$(BINARY_NAME)"; fi; \
```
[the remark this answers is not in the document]
mowies: done :)

Makefile
```diff
@@ -65,18 +69,67 @@ clean-tools:
...
+	ARCH=arm64
+endif
+
+EXT := tar.gz
+ifeq ($(OS), 'windows')
+	EXT='zip'
+endif
+
+# Construct binary name and URL
+BINARY_NAME := goreleaser-pro_$(OS)_$(ARCH).$(EXT)
```
evan-bradley, line 97: Nit: I think `ARCHIVE_NAME` would be a better name here since we're working with a tar/zip archive instead of the binary itself.

mowies: good point, done

joaopgrassi (approved):

PR #658, opened 2025-09-05
add k8sevents receiver, adapt tests

bacherfl:


config_examples/k8scombined-gateway.yaml
```diff
@@ -131,7 +127,7 @@ service:
         - otlphttp
     logs:
       receivers:
-        - k8sobjects
+        - k8s_events
       processors:
         - transform
```
odubajDT, line 132: I think we can leave out this part, as it won't be needed, since the k8seventsreceiver format the logs by default (compared to k8sobjectsreceiver)

PR #681, opened 2025-09-30
[extension/k8sleaderelector] introduce k8sleaderelector extension into the distribution

odubajDT:
Saves ~30s in test runtime

### DIFF
--- a/config_examples/k8scombined.yaml
+++ b/config_examples/k8scombined.yaml
@@ -97,8 +97,6 @@ processors:
         statements:
           - set(attributes["k8s.cluster.name"], "${env:CLUSTER_NAME}")
   k8sattributes:
-    filter:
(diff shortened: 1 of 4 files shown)


config_examples/k8scombined.yaml
```diff
@@ -68,8 +97,6 @@ processors:
         statements:
           - set(attributes["k8s.cluster.name"], "${env:CLUSTER_NAME}")
   k8sattributes:
-    filter:
-      node_from_env_var: K8S_NODE_NAME
```
mowies: can we then get rid of the env variable entirely?

mowies (approved):

config_examples/k8scombined.yaml
odubajDT: no, we still need it for kubeletstats receiver for node metrics

PR #692, opened 2025-10-07
[chore] fix host monitoring dashboard

odubajDT:
Screenshot 2025-10-07 at 10 11 41
Screenshot 2025-10-07 at 10 11 27


odubajDT:
> why did the dashboard json file get so much smaller? 😅

no idea, but works :D

PR #710, opened 2025-10-20
[kafka] add Kafka receiver and exporter to the distribution

odubajDT:


config_examples/kafka-exporter.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+receivers:
+  otlp:
+    protocols:
+      grpc:
+        endpoint: 0.0.0.0:4317
+
+exporters:
+  kafka:
+    brokers: ["redpanda.e2ekafka.svc.cluster.local:9092"]
```
bacherfl: should we rather set this via an env var?

PR #716, opened 2025-10-24
[chore] add integration test for gathering pod logs

bacherfl:


internal/testbed/integration/k8spodlogs/e2e_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	defer func() {
+		for _, obj := range append(collectorObjs, deploymentObj) {
+			require.NoErrorf(t, otelk8stest.DeleteObject(k8sClient, obj), "failed to delete object %s", obj.GetName())
+		}
+	}()
+
+	oteltest.WaitForLogs(t, 10, logsConsumer)
+
+	// search for the log of the test deployment in the logs received by the logsConsumer
```
odubajDT: if possible, let's try to do the comparison with plogtest.CompareLogs()

bacherfl: I did that previously, but since the resulting expected.yaml file included log entries from all pods in the cluster, this approach became too noisy and potentially flaky - therefore I thought this more targeted approach to specifically check for the expected log might be better in this case.

odubajDT (approved):

PR #721, opened 2025-10-30
[chore] Add setup for troubleshooting Kafka component issues

evan-bradley:


joaopgrassi (approved):

internal/testing-setups/kafka/go.mod
```diff
@@ -0,0 +1,52 @@
+module github.com/Dynatrace/dynatrace-otel-collector/internal/testing-setups/kafka
```
odubajDT: we can get rid of the `go.mod` and `go.sum` files and just execute 

```
GOBIN=$(pwd) go install github.com/open-telemetry/opentelemetry-collector-contrib/cmd/telemetrygen@latest
```

to install telemetrygen binary. Or is there some other reason why the `go.*` files are needed?

PR #927, opened 2026-04-07
[chore] Remove batch processor from docs

evan-bradley:
The batch processor is being deprecated in favor of the exporterhelper's sending queue. This PR removes it from our docs and switches batching to the new implementation.


internal/testbed/integration/self-monitoring/testdata/e2e/expected-check.yaml
odubajDT: these changes are very hard to read, can you point out what changed here? I would expect only the batch metrics were removed. Can you make the diff nicer and readable so also for the future we know what changed? Theoretically, only manual removal of those metrics can be sufficient, or we can use AI to make a nice diff, so the actual changes are visible.

odubajDT (changes requested):

internal/testbed/integration/self-monitoring/testdata/e2e/expected-check.yaml
evan-bradley: I agree they are difficult to read. I would like to switch to a different system, but there are efforts upstream to use Weaver to check for breaking changes in metrics, and I would prefer to align with those before we make further changes.

I have reorganized the diff a bit, you can now see the removed batch processor metrics on top. If you search for `name: ` in the file/diff you can see that no other metrics changed. It would also be possible to confirm this with a jq query as well.

internal/testbed/integration/self-monitoring/e2e_test.go
```diff
@@ -328,11 +328,6 @@ func Test_Selfmonitoring_checkMetrics(t *testing.T) {
 			"http.server.request.duration",
 			"http.client.request.body.size",
 			"http.client.request.duration",
-			"otelcol_processor_batch_batch_send_size",
```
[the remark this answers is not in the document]
evan-bradley: We do. The intent of this PR is to switch to use those in our self-monitoring dashboards.

odubajDT (approved):

internal/testbed/integration/self-monitoring/testdata/e2e/expected-check.yaml
odubajDT: thanks!

PR #953, opened 2026-04-16
[chore] Change collection interval for the most important host metrics

evan-bradley:
Changes the collection interval from 1m to 10s so our resolution is in line with similar collection by other tools.

### DIFF
--- a/config_examples/host-metrics.yaml
+++ b/config_examples/host-metrics.yaml
@@ -33,8 +33,8 @@ receivers:
       - type: move
         from: body.MESSAGE
         to: body.message
-  hostmetrics/1m:
-    collection_interval: 1m
+  hostmetrics/10s:
+    collection_interval: 10s
     scrapers:
       # CPU metrics
       cpu:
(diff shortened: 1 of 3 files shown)


config_examples/host-metrics.yaml
```diff
@@ -33,8 +33,8 @@ receivers:
       - type: move
         from: body.MESSAGE
         to: body.message
-  hostmetrics/1m:
-    collection_interval: 1m
+  hostmetrics/10s:
+    collection_interval: 10s
```
odubajDT, line 37: you need to remove the 1m overlay, it is not needed anymore https://github.com/Dynatrace/dynatrace-otel-collector/blob/main/internal/testbed/integration/hostmetrics/testdata/config-overlays/collection-interval-1m.yaml

Additionally, pipeline needs to be adapted, since you renamed the component

evan-bradley: Thanks. I pushed this yesterday just to see what would break and wasn't able to get back to it. Everything should be good now.

odubajDT (approved):
nits:
- we can rename the *1m variables in e2e_test.go - just to be consistent
- same for files which are named *1m*
- I think collection-interval-1m.yaml is not needed anymore - there is overlay to change to to 10s, but that we use now by default so no need for that I think

none of the above are blocking, but more nice to have and polishing - feel free to merge when you adapt those

PR #977, opened 2026-05-05
chore: exclude unneeded paths in pipeline triggers

mowies:


.github/workflows/ci.yaml
```diff
@@ -3,8 +3,20 @@ name: Continuous Integration
 on:
   push:
     branches: [main]
+    paths-ignore:
+      - "CHANGELOG.md"
+      - "docs/**"
+      - "**/*.md"
+      - "LICENSE"
+      - ".github/CODEOWNERS"
```
arminru: I think on main it would be fine if the CI ran nonetheless so that the latest commit always has binaries attached. Otherwise one would need to walk the commit history to find one that actually has a full build on it.

.github/workflows/ci.yaml
```diff
@@ -3,8 +3,20 @@ name: Continuous Integration
...
   pull_request:
     branches: [main]
+    paths-ignore:
+      - "CHANGELOG.md"
+      - "docs/**"
+      - "**/*.md"
+      - "LICENSE"
+      - ".github/CODEOWNERS"
```
arminru: Can this be de-duplicated into a central snippet for all workflow files?
Or would you rather keep it like that and we just hope it doesn't drift over time?

arminru: Unlikely that we change these two frequently, in case you want to keep it more concise.

arminru: @copilot Can this be de-duplicated into a central snippet for all workflow files?

.github/workflows/e2e.yaml
```diff
@@ -3,8 +3,16 @@ name: E2E-Test
...
+    paths-ignore:
+      - "CHANGELOG.md"
+      - "docs/**"
+      - "**/*.md"
   push:
     branches: [main]
+    paths-ignore:
+      - "CHANGELOG.md"
+      - "docs/**"
+      - "**/*.md"
```
arminru: @copilot remove the ignores from all push triggers in all workflow files touched in this PR, we only want it to be excluded on PRs.

.github/workflows/ci.yaml
```diff
@@ -5,6 +5,10 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+    paths-ignore:
+      - "CHANGELOG.md"
+      - "docs/**"
+      - "**/*.md"
```
arminru, line 11: @copilot will markdown PRs still be mergeable at all then, or will they be blocked on pending required merge checks that never run?

arminru (approved):

.github/workflows/ci.yaml
arminru: @copilot Check out https://github.com/Dynatrace/dynatrace-otel-collector/pull/978, there they don't show up as Skipped but are rather stuck at "Expected — Waiting for status to be reported | Required".

PR #980, opened 2026-05-06
chore: install cosign from binary

mowies:


odubajDT:
cosign already set up, closing

PR #982, opened 2026-05-06
chore: add large scale Prometheus examples

mowies:


config_examples/prometheus-large-scale/allocator.values.yaml
```diff
@@ -0,0 +1,44 @@
...
+      service_monitor_selector:
+        prometheus.dynatrace.com: "true"
+      pod_monitor_selector:
+        prometheus.dynatrace.com: "true"
+      scrape_config_selector:
+        prometheus.dynatrace.com: "true"
+
+#    filter_strategy: relabel-config
+#    config:
+#      scrape_configs: []
```
evan-bradley: If this is meant to be included, we should add a comment to this to explain when to uncomment it.

evan-bradley (commented):
Could we add a readme file to the new directory to briefly explain it?

config_examples/prometheus-large-scale/allocator.values.yaml
mowies: that is probably not needed at all. we full rely on the CRDs for TA configuration, i'll remove it for now

odubajDT (approved):

PR #1000, opened 2026-05-28
chore: add intergation test for large scale prometheus setup

odubajDT:


config_examples/prometheus-large-scale/selfmon-scraper.yaml
```diff
@@ -220,35 +230,37 @@ alternateConfig:
...
     extensions:
       - health_check
     pipelines:
       metrics:
         receivers:
+          - otlp
           - prometheus
         processors:
           - memory_limiter
+          - transform/selfmon_rename
```
odubajDT, line 252: we rename the `_` attributes from prometheus before enriching, so we don't have duplicated attributes in the end

internal/testbed/integration/prometheus-large-scale/e2e_test.go
```diff
@@ -0,0 +1,243 @@
...
+	expectedDataMetrics := []string{
+		"up",
+		"scrape_duration_seconds",
+		"scrape_samples_scraped",
+		"scrape_series_added",
+		"scrape_samples_post_metric_relabeling",
+	}
+
+	t.Log("Checking avalanche metrics from gateway...")
+	requireMetricsEventually(t, normalConsumer, expectedDataMetrics, 5*time.Minute)
```
pirgeo, line 133: So we are only making sure that

```
		"up",
		"scrape_duration_seconds",
		"scrape_samples_scraped",
		"scrape_series_added",
		"scrape_samples_post_metric_relabeling"
```

arrive at the data sink after some time. We don't check for any of the avalanche metric names, is that correct? I am surprised these metrics are not part of selfmon tbh

odubajDT: some of those are used in the dashboard, yes it's very weird those are part of the main data stream and we don't know the reasoning behind...

pirgeo: Ok. We don't check any other metrics from avalanche then?

odubajDT: no, because all other data in this stream are just mocked avalanche metrics to create load, nothing interesting to test there for us. For us is important that metrics are flowing and in this case check that metrics used in the dashboards are present^^. For other selfmon metrics we have other sinks and we check the metrics with the standard comparison mechanism we use everywhere

pirgeo (approved):

PR #1067, opened 2026-07-14
AI-158: introduce genainormalizerprocessor with OpenInference e2e tests

RealAnna:
## Summary

Adds `genainormalizerprocessor` to the Dynatrace collector distribution and covers it with an e2e integration test for the OpenInference instrumentation path.

- **`manifest.yaml`**: adds `genainormalizerprocessor` from `opentelemetry-collector-contrib`
- **`config-allcomponents.yaml`**: adds `gen_ai_normalizer` stanza to the all-components smoke test
- **`config_examples/genainormalizer-openinference.yaml`**: reference pipeline config (OTLP → gen_ai_normalizer → transform/response_model → DT exporter)
- **`internal/testbed/integration/genainormalizer/`**: new e2e test package
- **`.chloggen/`**: changelog entry (`new_component`)

## How the e2e test works

The test follows the same pattern as all other k8s e2e tests in this repo:

1. **Collector pod** — runs the Dynatrace OTel Collector with the `genainormalizer-openinference.yaml` config. The exporter endpoint is substituted to point at a local OTLP sink running in the test process.

2. **Crafted spans** — instead of a separate testapp pod or telemetrygen, the test constructs a `ptrace.Traces` with OpenInference `llm.*` attributes directly in Go and sends it to the collector via `kubectl port-forward`. This keeps the test self-contained with no external image to build or maintain.

The test asserts that `genainormalizerprocessor` correctly maps OpenInference attributes to `gen_ai.*` OTel semantic conventions:

| Input (OpenInference) | Output (gen_ai.*) |
|---|---|
| `llm.model_name` | `gen_ai.request.model`, `gen_ai.response.model` |
| `llm.token_count.prompt` | `gen_ai.usage.input_tokens` |
| `llm.token_count.completion` | `gen_ai.usage.output_tokens` |
| `llm.provider` | `gen_ai.provider.name` |
| `openinference.span.kind: LLM` | `gen_ai.operation.name: chat` |

Comparison uses a golden file; dynamic fields (trace/span IDs, timestamps) are ignored via `ptracetest` options.

## Test plan

- [x] `genainormalizer-e2e-test` CI job passes
- [x] All mapped `gen_ai.*` attributes are present and correct in the golden file
- [x] All-components smoke test passes with `gen_ai_normalizer` stanza

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/renderworkloads/main.go
```diff
@@ -124,7 +124,7 @@ func renderCollectorWorkloads(repoRoot, inRoot, outBase string, vars map[string]
 
 		if d.IsDir() {
 			switch filepath.Base(path) {
-			case ".git", "vendor":
+			case ".git", "vendor", "testapp":
```
RealAnna: skip security check for e2e test only applications

renovate.json
```diff
@@ -54,6 +54,9 @@
       "groupName": "GoReleaser Pro"
     }
   ],
+  "pip_requirements": {
```
RealAnna: this should make sure we actually capture changes in openinference  specs via libs

RealAnna:
kafka e2e test seems bogus 😢

internal/testbed/integration/genainormalizer/testapp/app.py
```diff
@@ -0,0 +1,96 @@
+import json
```
odubajDT: what is the motivation to have this testing app in this repository? it should be in a separate repo, since it has effectively nothing to do with collector and can be released on demand and the images can be pulled here. No need to effectively build the image for this app on every CI run.

internal/testbed/integration/genainormalizer/testdata/e2e/expected-traces.yaml
```diff
@@ -0,0 +1,340 @@
+resourceSpans:
```
odubajDT, line 1: we already migrated to a newer format of expected files, see here and please use the same https://github.com/Dynatrace/dynatrace-otel-collector/blob/main/internal/testbed/integration/k8scombined/testdata/e2e/expected-traces.yaml

internal/testbed/integration/genainormalizer/testapp/app.py
RealAnna: this is a mock app that needs to be updated everytime we have new dependencies but since we really need only one we could try to keep it here as a short lived solution. I now made a separate pipleine that will be triggered to release the new version so the test does not need to build at runtime.. of course test wont pass in this pr since no run has happened yet and I have no power to manually push a first image 🐱

RealAnna: should we separate the pipeline and test app pr at first and then continue with the test here? @odubajDT

odubajDT: would be probably the best, so we can verify everything works as expected

RealAnna: I asked @mowies to push an image to test from here first 🙏

.github/workflows/e2e.yaml
```diff
@@ -243,6 +243,22 @@ jobs:
           cd internal/testbed/integration/hostmetrics
           go test -v --tags=e2e
 
+  genainormalizer-e2e-test:
```
odubajDT: can't we run it as part of the standard testsuite? do we need a separate job for it?

RealAnna: leftover from previous build image will adapt 🦾

internal/testbed/integration/genainormalizer/e2e_test.go
```diff
@@ -0,0 +1,192 @@
...
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
+	require.NoError(t, err)
+	defer grpcConn.Close()
+
+	traceClient := ptraceotlp.NewGRPCClient(grpcConn)
+	traces := buildOpenInferenceTraces()
+	_, err = traceClient.Export(context.Background(), ptraceotlp.NewExportRequestFromTraces(traces))
+	require.NoError(t, err)
+
+	oteltest.WaitForTraces(t, 0, tracesConsumer)
```
odubajDT: aren't we waiting for 1?

internal/testbed/integration/genainormalizer/e2e_test.go
```diff
@@ -0,0 +1,192 @@
...
+		ptracetest.IgnoreScopeSpansOrder(),
+		ptracetest.IgnoreSpansOrder(),
+	}
+
+	const (
+		compareTimeout = 3 * time.Minute
+		compareTick    = 5 * time.Second
+	)
+
+	require.EventuallyWithT(t, func(tt *assert.CollectT) {
```
odubajDT: if we wait for exactly 1, we don't need eventually here, we can compare directly

internal/testbed/integration/genainormalizer/e2e_test.go
RealAnna: I think the function checks for traces > then number so needs to be 0 if it is exactly one

odubajDT (approved):

PR #1068, opened 2026-07-15
AI-158: add OpenLLMetry e2e test for genainormalizerprocessor

RealAnna:
## Summary

Extends the genainormalizerprocessor e2e test coverage with an OpenLLMetry instrumentation path, alongside the existing OpenInference test.

- **`config_examples/genainormalizer-openllmetry.yaml`**: reference pipeline config for OpenLLMetry (no `transform/response_model` needed — OpenLLMetry maps `llm.response.model` directly)
- **`TestE2E_GenAINormalizerProcessor_OpenLLMetry`**: new e2e test following the same pattern as the OpenInference test — crafted spans sent via `kubectl port-forward`, golden file comparison

## OpenLLMetry attribute coverage

The mock trace includes all mappable OpenLLMetry attributes, validated against the sdk-comparison-baseline generic profile:

| Input (OpenLLMetry) | Output (gen_ai.*) | Baseline |
|---|---|---|
| `llm.request.model` | `gen_ai.request.model` | AR-004 required |
| `llm.response.model` | `gen_ai.response.model` | AR-005 required |
| `llm.usage.prompt_tokens` | `gen_ai.usage.input_tokens` | AR-006 required |
| `llm.usage.completion_tokens` | `gen_ai.usage.output_tokens` | AR-007 required |
| `llm.request.type: chat` | `gen_ai.operation.name` | AR-008 recommended |
| `traceloop.entity.name` | `gen_ai.agent.name` | AR-010 recommended |
| `traceloop.entity.input` | `gen_ai.input.messages` | AR-011 optional |
| `traceloop.entity.output` | `gen_ai.output.messages` | AR-012 optional |
| `llm.request.temperature` | `gen_ai.request.temperature` | AR-042 optional |
| `gen_ai.system` (pass-through) | provider identity gate | AR-001/VR-001 |

Note: OpenLLMetry has no mapping for `gen_ai.provider.name` (AR-002) — the SDK sets `gen_ai.system` directly instead.

## Test plan

- [x] Golden file regenerated locally and committed
- [x] `genainormalizer` matrix job passes in CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.chloggen/genainormalizer-openllmetry.yaml
```diff
@@ -0,0 +1,18 @@
+# Use this changelog template to create an entry for release notes.
```
odubajDT: this does not need a dedicated changelog, it does not affect end user in any way, let's keep it as chore

internal/testbed/integration/genainormalizer/e2e_test.go
```diff
@@ -150,6 +150,124 @@ func TestE2E_GenAINormalizerProcessor_OpenInference(t *testing.T) {
 	require.NoError(t, ptracetest.CompareTraces(expectedTraces, tracesConsumer.AllTraces()[0], traceCompareOptions...))
 }
 
+// TestE2E_GenAINormalizerProcessor_OpenLLMetry verifies that the genainormalizer
+// processor correctly maps OpenLLMetry span attributes to gen_ai.* semantic conventions.
+// The test sends crafted spans with OpenLLMetry-style attributes directly to the
+// collector via kubectl port-forward, then compares the normalized output against a
+// golden file.
+func TestE2E_GenAINormalizerProcessor_OpenLLMetry(t *testing.T) {
```
odubajDT: do we need to set up the full collector deployment again? can't we just add this new testing use-case in the first one? I understand that the tests cannot affect each other, but this would be closer to real world, as collector would receive data from multiple sources and needs to deal with all of them at once, so we can make the test work like this

.chloggen/genainormalizer-openllmetry.yaml
RealAnna: We agreed with @mowies that is a good to have to show users we test also the OpenLLMetry path

internal/testbed/integration/genainormalizer/e2e_test.go
RealAnna: I wanted to make it cear which was failing but have nothing against this 🤷‍♀️

.chloggen/genainormalizer-openllmetry.yaml
mowies: however you like honestly. i thought it could be cool to show since it's so trendy :D

internal/testbed/integration/genainormalizer/e2e_test.go
odubajDT: you will still see which one fails, but we don't need to undeploy and redeploy the same setup

.chloggen/genainormalizer-openllmetry.yaml
odubajDT: we agreed to remove the chlog, the original one about introducing the component is enough. tests are not user-facing at all

internal/testbed/integration/genainormalizer/e2e_test.go
```diff
@@ -147,7 +141,17 @@ func TestE2E_GenAINormalizerProcessor_OpenInference(t *testing.T) {
 		ptracetest.IgnoreSpansOrder(),
 	}
 
-	require.NoError(t, ptracetest.CompareTraces(expectedTraces, tracesConsumer.AllTraces()[0], traceCompareOptions...))
+	// To regenerate golden files: uncomment the WriteTraces lines, run once, then revert.
+	// require.Nil(t, golden.WriteTraces(t, filepath.Join(testDir, "e2e", "expected-traces.yaml"), tracesConsumer.AllTraces()[0]))
+	// require.Nil(t, golden.WriteTraces(t, filepath.Join(testDir, "e2e", "expected-traces-openllmetry.yaml"), tracesConsumer.AllTraces()[1]))
+
+	expectedOI, err := golden.ReadTraces(filepath.Join(testDir, "e2e", "expected-traces.yaml"))
```
odubajDT: are we sure that these spans always arrive in the right order? It can cause flakyness if not

odubajDT (approved):

PR #1082, opened 2026-07-21
chore: Add license file headers and tooling for them

mowies:
For reviewers: There is one commit with all the tooling, and one commit with all the license headers.


joaopgrassi (approved):

callegarimattia (approved):
just out of curiosity: why some path are excluded from the the license check? for instance, the `internal/tools/` one

mowies:
@callegarimattia i added the tools folder and also the dockerfile now. thanks for the heads-up

PR #1090, opened 2026-07-24
feat: add bearertokenauth extension to the distribution

odubajDT:


internal/testbed/integration/bearertokenauth/e2e_test.go
```diff
@@ -0,0 +1,141 @@
...
+			require.NoErrorf(t, otelk8stest.DeleteObject(k8sClient, obj), "failed to delete object %s", obj.GetName())
+		}
+	}()
+
+	for _, info := range telemetryGenObjInfos {
+		otelk8stest.WaitForTelemetryGenToStart(t, k8sClient, info.Namespace, info.PodLabelSelectors, info.Workload, info.DataType)
+	}
+
+	// If bearer token auth worked end-to-end, traces will arrive at the sink.
+	wantEntries := 5
```
callegarimattia, line 139: why specifically 5?

callegarimattia (approved):

internal/testbed/integration/bearertokenauth/e2e_test.go
odubajDT: can be also 1 or 10, it does not matter that much, telemetrygen is sending data periodically, 1 seems too few to verify we have data arriving continuously, 10 will already make the test a bit slower without additional value. 5 seems like a sweet spot :)

callegarimattia: gotcha!

PR #1099, opened 2026-07-29
chore: make syslog RFC5424 example tests deterministic across platforms

callegarimattia:
## Problem

`TestSyslog_WithF5Receiver` and `TestSyslog_WithHostReceiver`
(`internal/testbed/integration/config_examples_test.go`) pass on CI
(`linux/amd64`) but fail when run natively on `darwin/arm64`.

### Root cause

The tests seed the log timestamp from `time.Now()` and the testbed syslog
sender formats it with `time.RFC3339Nano`. The RFC5424 parser
(`leodido/go-syslog`, `RFC3339MICRO`) accepts **at most 6** fractional-second
digits.

- **Linux CI:** nanosecond clock → 7-9 digits → the parser **rejects** the
  timestamp → the syslog parse fails silently → the receiver attaches **none**
  of the parsed fields → only the 4 static `add` attributes remain → matches the
  expected record → **green by accident**.
- **darwin/arm64:** microsecond-granular clock → ≤6 digits → the timestamp is
  valid → the parse **succeeds** → `facility`, `priority`, `hostname`,
  `message`, `structured_data`, `version`, severity are all populated → the
  exact-match validator fails.

In other words, the tests only pass on CI because RFC5424 parsing never actually
runs there. They validate nothing about syslog parsing today.

## This PR (two commits, on purpose)

1. **`test: normalize ... timestamp to microsecond precision`** — truncates the
   timestamp to µs so parsing is deterministic on every platform, but
   **intentionally leaves the expectations unchanged**. This makes CI
   (`linux/amd64`) go **red**, proving the receiver really does populate the
   RFC5424 fields once the timestamp is valid — i.e. the previous green was an
   artifact of nanosecond timestamps.
2. **`fix(test): assert the RFC5424 fields ...`** _(incoming)_ — updates the
   expected records to include the parsed fields, severity, and parsed
   timestamp, so the tests are green on both `linux/amd64` and `darwin/arm64`
   and actually exercise the parser.

Draft until commit 2 is pushed and CI is green.

### DIFF
--- a/internal/testbed/integration/config_examples_test.go
+++ b/internal/testbed/integration/config_examples_test.go
@@ -598,7 +598,11 @@ func TestSyslog_WithF5Receiver(t *testing.T) {
 	expectedLogsData := plog.NewLogs()
 	expectedLogs := expectedLogsData.ResourceLogs().AppendEmpty().ScopeLogs().AppendEmpty().LogRecords()
 
-	timestamp := time.Now()
+	// Truncate to microseconds so the RFC5424 timestamp the sender emits (formatted with
+	// time.RFC3339Nano) never exceeds 6 fractional-second digits. On Linux the nanosecond
+	// clock produces 7-9 digits, which the syslog parser rejects, causing the parse to fail
+	// silently and hiding the fields the receiver would otherwise populate.
+	timestamp := time.Now().Truncate(time.Microsecond)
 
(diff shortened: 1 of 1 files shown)


pichlermarc (approved):
Looks good, thanks for chasing that down + fixing this 🙌

odubajDT (approved):

PR #1101, opened 2026-07-29
chore: add local e2e make targets

callegarimattia:
## Summary

- Adds `make e2e-build` to cross-compile a `linux/ARCH` binary and build both Docker images (`dynatrace-otel-collector:e2e-test` and `-journald:e2e-test`) locally
- Adds `make e2e-test SUITE=<suite>` for a full single-suite lifecycle: cluster create → image load → swap setup → test → cluster teardown (always, via trap)
- Adds `make e2e` to build once and run all suites in the CI matrix sequentially
- Adds findings doc covering port conflict analysis, resource envelope, and suite timings for future parallelism work

The swap setup step (`dd`/`mkswap`/`swapon` on the Kind node's ext4 `/var` partition) is required on macOS because the Rancher Desktop VM has no swap by default, which would otherwise cause `system.paging.usage` assertions to fail.

## Test plan

- [x] `make e2e-build` — builds linux/arm64 binary and both images
- [x] `make e2e-test SUITE=hostmetrics` — full lifecycle, test passes
- [x] `make e2e` — all 20 suites green, ~33 min total


Makefile
```diff
@@ -102,6 +102,58 @@ for-all-target: $(INTERNAL_MODS)
...
+		-o $(E2E_DOCKER_CTX)/dynatrace-otel-collector .
+	docker build -t $(E2E_IMAGE) -f Dockerfile $(E2E_DOCKER_CTX)/
+	docker build -t $(E2E_IMAGE_JRN) -f Dockerfile-journald $(E2E_DOCKER_CTX)/
+
+e2e-test:
+	@[ "$(SKIP_E2E_BUILD)" = "true" ] || $(MAKE) e2e-build
+	@[ -n "$(SUITE)" ] || { echo "Error: SUITE is required. Usage: make e2e-test SUITE=<suite>"; exit 1; }
+	@trap 'kind delete cluster --name $(E2E_CLUSTER_NAME) 2>/dev/null; true' EXIT; \
+	kind delete cluster --name $(E2E_CLUSTER_NAME) 2>/dev/null || true; \
+	kind create cluster \
```
odubajDT, line 138: I would avoid creating cluster here, users might already have a cluster set up and this can mess up their setup. Also, kind is not the only possibility how to spin up a cluster (it's also allowed to be used only on exception), so I would completely avoid this. Only possibility would be to have it in a separate Makefile command and users can use it if they want

Makefile
```diff
@@ -102,6 +102,58 @@ for-all-target: $(INTERNAL_MODS)
...
+		--name $(E2E_CLUSTER_NAME) \
+		--image kindest/node:$(E2E_K8S_VERSION) \
+		--kubeconfig $(E2E_KUBECONFIG) \
+		--config $(E2E_CLUSTER_CONFIG); \
+	kind load docker-image $(E2E_IMAGE) --name $(E2E_CLUSTER_NAME); \
+	kind load docker-image $(E2E_IMAGE_JRN) --name $(E2E_CLUSTER_NAME); \
+	docker exec $(E2E_CLUSTER_NAME)-control-plane bash -c \
+		'dd if=/dev/zero of=/var/swapfile bs=1M count=512 status=none && chmod 600 /var/swapfile && mkswap /var/swapfile && swapon /var/swapfile' 2>/dev/null || true; \
+	cd internal/testbed/integration/$(SUITE) && \
+	KUBECONFIG=$(E2E_KUBECONFIG) go test -v --tags=e2e
```
odubajDT, line 148: Running the full suite will probably fail, since each test requires quite some resources and it will take very long to run it sequentially. Paralelism is not an option here as well. I am not sure if we want to have it here

PR #1127, opened 2026-08-06
[chore] add release ingestion via API-GW

dominikeinkemmer:


.github/actions/send-release-notification/action.yml
```diff
@@ -0,0 +1,42 @@
...
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Configure AWS credentials for release ingestion
+      uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6.2.3
+      with:
+        role-to-assume: ${{ inputs.role_arn }}
+        aws-region: eu-central-1
```
mowies: this could be a parameter or an env var since it's used here and also inside the python script

.github/actions/send-release-notification/send_release_notification.py
```diff
@@ -0,0 +1,53 @@
...
+aws_req = AWSRequest(method="POST", url=url, data=body, headers={
+    "Content-Type":            "application/json",
+    "X-Release-Signature-256": signature,
+    "X-Release-Delivery-Id":   delivery_id,
+})
+SigV4Auth(boto3.Session().get_credentials(), "execute-api", "eu-central-1").add_auth(aws_req)
+
+req = urllib.request.Request(url, data=body, headers=dict(aws_req.headers), method="POST")
+with urllib.request.urlopen(req) as resp:
+    print(f"Ingestion API response: {resp.status}")
```
mowies: this would probably benefit from a retry logic to make sure we overcome random http errors and network glitches. also a timeout would also be good to not keep this running forever if it somehow hangs

.github/actions/send-release-notification/send_release_notification.py
```diff
@@ -0,0 +1,53 @@
...
+    "artifacts": [{
+        "type":       "docker-image",
+        "name":       component,
+        "source_ref": source_ref,
+        "digest":     digest,
+    }],
+}
+
+body        = json.dumps(payload, separators=(",", ":")).encode()
+delivery_id = str(uuid.uuid4())
```
mowies: is there some possibility to make this id generation deterministic? if we need to re-try the release run for some reason this will generate a completely new ID

.github/actions/send-release-notification/send_release_notification.py
```diff
@@ -0,0 +1,53 @@
...
+version = tag.lstrip("v")
+repo    = os.environ["GITHUB_REPOSITORY"]
+
+ghcr_prefix = os.environ["GHCR_PREFIX"]
+target_ref  = f"{ghcr_prefix}:{version}"
+
+source_ref = digest = None
+with open(os.environ["DIGESTS_FILE"]) as f:
+    for line in f:
+        ref, dgst = line.strip().split()
```
mowies: the digests.txt file is of the following format:
```
digest1  img1
digest2  img2
digest3  img3
```
so the columns are reversed here.
```suggestion
        dgst, ref = line.strip().split()
```

mowies (approved):
LGTM

PR #1138, opened 2026-08-11
feat: add oidcauth extension to distribution

odubajDT:


internal/testbed/integration/bearertokenauth/e2e_test.go
```diff
@@ -236,3 +242,186 @@ func TestE2E_BearerTokenAuthReceiver(t *testing.T) {
...
+	require.NoError(t, err)
+
+	nsFile := filepath.Join(testDir, "namespace.yaml")
+	buf, err := os.ReadFile(nsFile)
+	require.NoErrorf(t, err, "failed to read namespace object file %s", nsFile)
+	nsObj, err := otelk8stest.CreateObject(k8sClient, buf)
+	require.NoErrorf(t, err, "failed to create k8s namespace from file %s", nsFile)
+
+	testNs := nsObj.GetName()
+	defer func() {
```
callegarimattia, line 268: I think `t.Cleanup()` is better suited for teardowns

internal/testbed/integration/bearertokenauth/e2e_test.go
```diff
@@ -236,3 +242,186 @@ func TestE2E_BearerTokenAuthReceiver(t *testing.T) {
...
+	}()
+
+	// Overlay that redirects each bad sender's exporter to the verifier.
+	endpointOverlay := fmt.Sprintf(k8stest.MustRead(t, filepath.Join(testDir, "config-overlays", "sender-endpoint.yaml")), verifierSvcEndpoint)
+
+	// Three negative cases.  Each uses a distinct rejection path in oidcauthextension:
+	//   wrong-audience  — valid cluster-issued JWT, but aud != "dynatrace-wif-test"
+	//   invalid-token   — static string that is not a parseable JWT (no issuer extraction)
+	//   no-token        — no Authorization header at all
+	cases := []struct {
```
callegarimattia, line 319: Just for me to understand: is there a reason why you decided to do the three cases together instead of doing a t.Test() for each in a table driven test?

internal/testbed/integration/bearertokenauth/e2e_test.go
callegarimattia, line 1: Is the positive path missing? where the oidc is actually validated? I couldn't find it :C

config_examples/README.md
```diff
@@ -30,6 +30,8 @@ Dynatrace distribution of the OpenTelemetry Collector.
 - [Host Metrics Receiver](host-metrics.yaml)
 - [Dynatrace Resource Detector](resource-detection.yaml)
 - [Large Scale Prometheus Scraping](./prometheus-large-scale)
+- [Bearer Token Auth Receiver](bearertokenauth-receiver.yaml)
```
callegarimattia, line 33: was this one missing from before?

internal/testbed/integration/bearertokenauth/e2e_test.go
odubajDT: no explicit reason, but the use-cases do not affect each other in any way and we save some code, since the requirements for all 3 use-cases for the test to be successful are:

- no data arrive on sink
- error log messages on the sender side
- error log message on the receiver side

If we can check all^^ without any interference, we're good I guess.

callegarimattia (commented):
Overall lgtm, just left some questions for me and a couple of nitpicks

internal/testbed/integration/bearertokenauth/e2e_test.go
odubajDT: good point, worth a try

internal/testbed/integration/bearertokenauth/e2e_test.go
odubajDT: it was already present - TestE2E_BearerTokenAuth, just adapted the image to use the dynatrace-otel-collector one and the config from config_examples

odubajDT: https://github.com/Dynatrace/dynatrace-otel-collector/pull/1138/changes#diff-37c632664700b0d02ff019e776addb31da94d79acb0fcc888a3175e71d86f0e1R71

config_examples/README.md
odubajDT: yep, we missed it before, added it now

internal/testbed/integration/bearertokenauth/e2e_test.go
odubajDT: should be completely the same functional-wise, but since we're using defer() in all tests, we can rather do a follow-up to adapt it everywhere than only using it just here. Just for consistency.

callegarimattia (approved):

PR #1153, opened 2026-08-20
[chore] Add OpenSSF Scorecard

Kenyatta-forbes:
Repo has been identified as a strategic repo, OSPO will add the Open SSF Scorecard to the repo in the README


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,45 @@
+name: OpenSSF Scorecard
+
+on:
+  schedule:
+    - cron: "17 3 * * 1"
```
mowies, line 5: oddly specific time? 😄

mowies (changes requested):
please add hashes to all action versions in the workflow like is done with all other actions in the repo

.github/workflows/scorecard.yml
Kenyatta-forbes: 😄 Fair callout. The odd minute is intentional. I picked a non-round time to avoid having every scheduled workflow fire exactly on the hour. Happy to change it if there’s a preferred org convention for scheduled jobs.

mowies (changes requested):
please add hashes to all action versions in the workflow like is done with all other actions in the repo

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,45 @@
...
+      security-events: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Run OpenSSF Scorecard
+        uses: ossf/scorecard-action@v2.4.4
```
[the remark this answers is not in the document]
Kenyatta-forbes: Updated

mowies (approved):

PR #1154, opened 2026-08-20
Update README.md

Kenyatta-forbes:
adding OpenSSF Scorecard badge


mowies (changes requested):
blocking this until we reviewed the results of #1153

PR #1165, opened 2026-08-25
[chore] fix(ci): make required job run and succeeed on docs change

callegarimattia:


.github/workflows/ci.yaml
```diff
@@ -5,19 +5,39 @@ on:
...
-      - "CHANGELOG.md"
-      - "docs/**"
   workflow_dispatch:
 
 env:
   # renovate: datasource=golang-version depName=go
   GO_VERSION: "1.27.0"
 
 jobs:
+  detect-changes:
```
odubajDT, line 15: this is the same block over and over again, let's put it in a separate action and parametrize it (if needed), for example you can look at the create-cluster action we have here

.github/workflows/detect-changes.yaml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+      - id: check
+        run: |
+          if [[ "${{ github.event_name }}" != "pull_request" ]]; then
+            echo "skip=false" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
```
odubajDT: could this be moved above to the workflow conditions? something like

```
on:
    pull_request:
```

.github/workflows/detect-changes.yaml
```diff
@@ -0,0 +1,30 @@
...
+      skip: ${{ steps.check.outputs.skip || 'false' }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+      - id: check
+        if: github.event_name == 'pull_request'
+        run: |
+          if git diff --name-only origin/${{ github.base_ref }}...HEAD \
+              | grep -qvE '^(docs/|CHANGELOG\.md|.*\.md$)'; then
```
mowies: nitpick: would be nice if the list of exceptions can be configured in a nice, maybe list-resembling way

mowies (approved):

.github/workflows/detect-changes.yaml
callegarimattia: good point. committed :)

.github/workflows/detect-changes.yaml
```diff
@@ -0,0 +1,36 @@
...
+    outputs:
+      skip: ${{ steps.check.outputs.skip || 'false' }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+      - id: check
+        if: github.event_name == 'pull_request'
+        env:
+          SKIP_PATTERNS: |
```
callegarimattia, line 25: @mowies this should be a bit easier to modify/read.

mowies (approved):

odubajDT (approved):

